PR #6, opened 2022-06-09
#5: Prepared skeleton

### LINKED ISSUE #5: Prepare the skeleton of the project
- setup script language container installation
- add deployment scripts
- add deployment tests
- prepare test fixtures

umitbuyuksahin:
Fixes #5


exasol_transformers_extension/deployment/language_container_deployer.py
```diff
@@ -0,0 +1,111 @@
...
+        path_in_udf_without_bucksts = Path(*path_in_udf.parts[2:])
+        new_language_alias_definition = \
+            f"{self._language_alias}=localzmq+protobuf:///" \
+            f"{path_in_udf_without_bucksts}?lang=python#" \
+            f"{path_in_udf}/exaudf/exaudfclient_py3"
+        new_definitions = other_definitions + [new_language_alias_definition]
+        new_definitions_str = " ".join(new_definitions)
+        return new_definitions_str
+
+    def check_if_requested_language_alias_already_exists(
```
tkilias, line 74:
```suggestion
    def _check_if_requested_language_alias_already_exists(
```

exasol_transformers_extension/resources/templates/model_downloader_udf.jinja.sql
```diff
@@ -0,0 +1,9 @@
+CREATE OR REPLACE {{ language_alias }} SCALAR SCRIPT "TE_MODEL_DOWNLOADER_UDF"(
+    model_name VARCHAR(2000000),
+    path_in_the_bucket VARCHAR(2000000),
```
tkilias, line 3: I wouldn't specify the path in bucket via an extra parameter, I would push this all into the bfs_con, because I think, we can assume the following scenario, an admin creates the bfs connection object and defines where the cache should be and the user get access granted for this connection.

exasol_transformers_extension/udfs/model_downloader_udf.py
```diff
@@ -0,0 +1,8 @@
+
+
+class ModelDownloader:
+    def __init__(self, exa):
+        self.exa = exa
+
+    def run(self, ctx) -> None:
+        pass
```
tkilias, line 8: If I add a mock without implementation, I usually raise NotImplementedError, this ensures that tests fail for this function fail, until I have it implemented.

language_container/exaslct_scripts/exaslct.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+SCRIPT_DIR="$(dirname "$(readlink -f "${BASH_SOURCE[0]}")")"
+
+EXASLCT_GIT_REF="0.9.0"
```
tkilias, line 7: We are already at version 0.14.0. Please have a look, how to update. https://github.com/exasol/script-languages-container-tool#starter-scripts. I suggest you add the container-tool package to the dev dependencies.

tkilias (changes requested):

tkilias (approved):

PR #7, opened 2022-06-10
#4: Added model downloader UDF

### LINKED ISSUE #4: Add model downloaderUDF
# Background

- In order to use the pre-trained models, we need to download them using transformers api
- The downloaded models should be stored into bucketfs to use in prediction operations.

# Acceptance Criteria
- An UDF
  - downloads the specified model into tmp folder
  - uploads the downloaded model into bucketfs
- The UDF should take bucketfs conection, and bucketfs path to store model as input parameters
- Add unit and integration tests

umitbuyuksahin:
Fixes #4


exasol_transformers_extension/udfs/model_downloader_udf.py
```diff
@@ -1,8 +1,40 @@
...
+            user=bfs_conn_obj.user,
+            pwd=bfs_conn_obj.password)
+
+        with tempfile.TemporaryDirectory() as tmpdirname:
+            # download model into tmp folder
+            self.downloader_method.from_pretrained(
+                model_name, cache_dir=tmpdirname)
+
+            # upload the downloaded model files into bucketfs
+            for tmp_file_name in os.listdir(tmpdirname):
```
tkilias, line 33: do we only have files or also sub directories?

exasol_transformers_extension/udfs/model_downloader_udf.py
```diff
@@ -1,8 +1,40 @@
+import os
+import tempfile
+import transformers
+from pathlib import PurePosixPath
+from exasol_bucketfs_utils_python.bucketfs_factory import BucketFSFactory
 
 
 class ModelDownloader:
-    def __init__(self, exa):
+    def __init__(self, exa, downloader_method=transformers.AutoModel):
```
tkilias, line 9: does it download different models, when you are using AutoModel and AutoTokenizer

exasol_transformers_extension/udfs/model_downloader_udf.py
tkilias: maybe document the structure of the cache, also for the case of multiple models

tests/unit_tests/udfs/test_model_downloader_udf.py
```diff
@@ -0,0 +1,81 @@
...
+        exa = MockExaEnvironment(
+            metadata=meta,
+            connections={BFS_CONN_NAME: bucketfs_connection})
+        input_data = (
+            MODEL_NAME,
+            BFS_CONN_NAME)
+        result = executor.run([Group([input_data])], exa)
+
+        relative_model_path = MODEL_NAME.replace('-', '_')
+        full_model_path = os.path.join(path, relative_model_path)
```
tkilias, line 74: pathlib should be used instead of os.path

tkilias (changes requested):

exasol_transformers_extension/udfs/model_downloader_udf.py
umitbuyuksahin: we have only files. 

These files are uploaded to bucketfs right after downloading to tmp directory. For this reason, I did not see the need to make an arrangement in the tmp directory. But there are separate directories for each model in bucketfs.

exasol_transformers_extension/udfs/model_downloader_udf.py
umitbuyuksahin: AutoModel and AutoTokenizer seems different models. The AutoTokenizer model should be additionally saved   for each model.

tests/fixtures/setup_database_fixture.py
```diff
@@ -0,0 +1,41 @@
...
+    ScriptsDeployer.run(
+        dsn=db_params.address(),
+        user=db_params.user,
+        password=db_params.password,
+        schema=schema_name,
+        language_alias=language_alias
+    )
+
+
+def _create_bucketfs_connection(db_conn) -> None:
```
tkilias, line 28: I would maybe use the bucket_config and maybe use https://github.com/exasol/bucketfs-utils-python/blob/main/exasol_bucketfs_utils_python/bucketfs_utils.py#L107

tests/unit_tests/udfs/test_model_downloader_udf.py
```diff
@@ -0,0 +1,103 @@
...
+BFS_CONN_NAME = "test_bfs_conn_name"
+MODEL_FILE_DATA_MAP = {
+    "model_file1.txt": "Sample data in model_file1.txt",
+    "model_file2.txt": "Sample data in model_file1.txt"}
+TOKENIZER_FILE_DATA_MAP = {
+    "tokenizer_file1.txt": "Sample data in tokenizer_file1.txt",
+    "tokenizer_file2.txt": "Sample data in tokenizer_file1.txt"}
+
+
+def udf_wrapper():
```
tkilias, line 22: I would suggest, to add another out of db test, which actually downloads from hugging face, because this way, you don't need to debug in the UDF, if something goes wrong with the transformer lib.

tests/integration_tests/test_model_downloader_udf_implementation.py
```diff
@@ -0,0 +1,75 @@
...
+        user=bucketfs_params.user,
+        password=bucketfs_params.password)
+    exa = ExaEnvironment({bucketfs_conn_name: bucketfs_connection})
+
+    try:
+        model_downloader = ModelDownloader(exa)
+        model_downloader .run(ctx)
+
+        # assertions
+        path_in_the_bucket = str(Path("container", f"{model_path}"))
```
tkilias, line 62: why do you store your models under container

tkilias: path in bucket needs to be defined outside if the try block, because you are using it in the finally

tests/integration_tests/test_model_downloader_udf_implementation.py
```diff
@@ -0,0 +1,75 @@
...
+    bucketfs_conn_name = "bucketfs_connection"
+    model_name = 'bert-base-uncased'
+    model_path = model_name.replace("-", "_")
+
+    ctx = Context(
+        model_name,
+        bucketfs_conn_name
+    )
+
+    bucketfs_connection = Connection(
```
tkilias, line 51: I would replace the real bucketfs with the Mock, see BucketFSLocationFactory, because then you can run this without the DB and as unit test

tests/integration_tests/test_model_downloader_udf_implementation.py
```diff
@@ -0,0 +1,75 @@
...
+        assert ctx.get_emitted()[0][0] == model_path and files
+    finally:
+        # revert, delete downloaded model files
+        try:
+            files = list_files.list_files_in_bucketfs(
+                bucket_config, path_in_the_bucket)
+            for file_ in files:
+                delete.delete_file_in_bucketfs(
+                    bucket_config, str(Path(path_in_the_bucket, file_)))
+        except Exception as exc:
```
tkilias, line 74: Move the try catch into the for loop, such that you can continue to delete the other files after an error

tkilias (changes requested):

tests/integration_tests/test_model_downloader_udf_implementation.py
umitbuyuksahin: agree :+1:

tkilias:
@umitbuyuksahin We get the size issue, because the language container is 2.2 GB and we have around 4 Copies of it. The docker images we build, the exported container, the upload container in the bucketfs and the extracted container in the bucketfs. I don't think, we can get enough space in Github Actions and should move the build to AWS Code Build.

tests/integration_tests/without_db/test_download_pretrained_model.py
```diff
@@ -0,0 +1,22 @@
+import json
+import tempfile
+import transformers
+from pathlib import Path
+
+
+def test_download_pretrained_model():
```
tkilias, line 7: Add a comment, regarding our intention with this test which is "Checking expected behavior of the 3rd party api"

tkilias (approved):

PR #10, opened 2022-07-07
#9: Created sequence classification udf

### LINKED ISSUE #9: Create Sequence Classification UDF
Create an UDF for sequence text classification (please see [sequence-classification](https://huggingface.co/docs/transformers/task_summary#sequence-classification)) 
- takes variadic  inputs.
- returns softmax logits
- ex: transformers_sequence_classification(string, ..., string) -> softmax_logits

umitbuyuksahin:
Fixes #9


exasol_transformers_extension/udfs/sequence_classification_udf.py
```diff
@@ -0,0 +1,43 @@
...
+        model_path = bucketfs_operations.get_model_path(model_name)
+
+        # create bucketfs location
+        bfs_conn_obj = self.exa.get_connection(bfs_conn)
+        bucketfs_location = bucketfs_operations.create_bucketfs_location(
+            bfs_conn_obj)
+
+        with tempfile.TemporaryDirectory() as tmpdir_name:
+            # download the model files from bucketfs
+            bucketfs_operations.download_model_files_from_bucketfs(
```
tkilias, line 28: We don't want to download or read the files from the bucketfs, because they are large. We want to specify the mounted local BucketFs path as cache dir for the from_pretrained functions

exasol_transformers_extension/udfs/sequence_classification_udf.py
```diff
@@ -0,0 +1,43 @@
...
+                tmpdir_name, model_path, bucketfs_location)
+
+            # load models
+            model = self.base_model.from_pretrained(
+                model_name, cache_dir=tmpdir_name)
+            tokenizer = self.tokenizer.from_pretrained(
+                model_name, cache_dir=tmpdir_name)
+
+            for i in range(2, self.exa.meta.input_column_count):
+                tokens = tokenizer(ctx[i], return_tensors="pt")
```
tkilias, line 38: The variadic parameters of one row go all together in the tokenizer. So, one tokenizer call per row.

tests/unit_tests/udfs/test_sequence_classification.py
```diff
@@ -0,0 +1,112 @@
...
+            connections={BFS_CONN_NAME: bucketfs_connection})
+
+        input_data = (
+            BFS_CONN_NAME,
+            str(model_path),  # note that model name is used as model path TODO!
+            "Test text 1",
+            "Test text 2",
+            "Test text 3"
+        )
+        result = executor.run([Group([input_data])], exa)
```
tkilias, line 100: Use multiple rows for testing

exasol_transformers_extension/resources/templates/sequence_classification_udf.jinja.sql
```diff
@@ -0,0 +1,5 @@
+CREATE OR REPLACE {{ language_alias }} SCALAR SCRIPT "TE_SEQUENCE_CLASSIFICATION_UDF"(...)EMITS  (outputs VARCHAR(2000000)) AS
```
tkilias, line 1: Don't use a scalar UDF, because it is difficult to control the number of instances for them. You need also add the iterating over the rows to the UDF.

tkilias (changes requested):

exasol_transformers_extension/udfs/sequence_classification_udf.py
umitbuyuksahin: tokenizer does not receive variadic inputs. It gets either string or list of string as input (giving the list of string means the pretokenized the text.). Please see the [API](https://huggingface.co/docs/transformers/main_classes/tokenizer)

exasol_transformers_extension/udfs/sequence_classification_single_text_udf.py
```diff
@@ -0,0 +1,44 @@
...
+                 cache_dir=None,
+                 base_model=transformers.AutoModelForSequenceClassification,
+                 tokenizer=transformers.AutoTokenizer):
+        self.exa = exa
+        self.cache_dir = cache_dir
+        self.base_model = base_model
+        self.tokenizer = tokenizer
+
+    def run(self, ctx):
+        bucketfs_conn = ctx.bucketfs_conn
```
tkilias, line 18: You need to iterate over the input and if possible we want to use batch prediction, but careful, the model could change. It is probably a good idea to add a order by on bucketfsconn and model parameters of the UDF.

exasol_transformers_extension/udfs/sequence_classification_single_text_udf.py
```diff
@@ -0,0 +1,44 @@
...
+        text = ctx.text
+        model_name = ctx.model_name
+
+        # get bucketfs location
+        bucketfs_conn_obj = self.exa.get_connection(bucketfs_conn)
+        bucketfs_location = bucketfs_operations.create_bucketfs_location(
+            bucketfs_conn_obj)
+
+        # get cache directory
```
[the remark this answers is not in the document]
umitbuyuksahin: if  the cache_dir is not given, we shouldn't fail because we get it from model_name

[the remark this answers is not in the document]
tkilias: This means we can remove the cache_dir, because in the downloader we expect the bucketfs connection gives us the cache_dir root, and we then store the model under its name. Alternatively, we use it in DownloaderUDF and PredictionUDF, to allow the user to specify subdirectories for the bucketfs con. The latter is probably a good idea, because the user might get a single connection from the admin and needs to use it for different things.

exasol_transformers_extension/udfs/sequence_classification_single_text_udf.py
```diff
@@ -7,28 +8,70 @@ class SequenceClassificationSingleText:
...
+            batch_df = ctx.get_dataframe(self.bacth_size)
+            if batch_df is None:
+                break
+
+            result_df = self.get_batched_predictions(batch_df)
+            ctx.emit(result_df)
+
+    def get_batched_predictions(self, batch_df):
+        result_df_list = []
+        for model_name in batch_df['model_name'].unique():
```
tkilias, line 34: in which order do you get the model_name, it would be good, if it is the same order as in the original data_frame

exasol_transformers_extension/udfs/sequence_classification_single_text_udf.py
```diff
@@ -7,28 +8,70 @@ class SequenceClassificationSingleText:
...
+
+    def model_prediction(self, model_df):
+        bucketfs_conn = model_df['bucketfs_conn'].iloc[0]
+        model_name = model_df['model_name'].iloc[0]
+
+        model_preds = []
+        for ix, row in model_df.iterrows():
+            text_data = row['text_data']
+            tokens = self.last_loaded_tokenizer(text_data, return_tensors="pt")
+            logits = self.last_loaded_model(**tokens).logits
```
tkilias, line 64: can you check, if tokenizer and/or model have batch capabilities. Here a first hint https://huggingface.co/course/chapter2/5?fw=tf. Another thing, for later, maybe think about, how you can refactor this into using two Threads, one thread which interacts with the UDF context and one who does the prediction. The reason for this refactoring is performance. The prediction will run eventually on GPU and during that the CPU is free to request new data from the DB or emit results.

exasol_transformers_extension/resources/templates/sequence_classification_single_text_udf.jinja.sql
```diff
@@ -2,7 +2,13 @@ CREATE OR REPLACE {{ language_alias }} SET SCRIPT "TE_SEQUENCE_CLASSIFICATION_SI
     bucketfs_conn VARCHAR(2000000),
     model_name VARCHAR(2000000),
     text_data VARCHAR(2000000)
-)EMITS  (...) AS
+    ORDER BY model_name
+)EMITS (
+    bucketfs_conn VARCHAR(2000000),
+    model_name VARCHAR(2000000),
+    text_data VARCHAR(2000000),
+    labels VARCHAR(2000000),
```
tkilias, line 10: label

exasol_transformers_extension/resources/templates/sequence_classification_single_text_udf.jinja.sql
```diff
@@ -2,7 +2,13 @@ CREATE OR REPLACE {{ language_alias }} SET SCRIPT "TE_SEQUENCE_CLASSIFICATION_SI
...
+    scores DOUBLE ) AS
```
tkilias, line 11: score

tkilias (changes requested):

exasol_transformers_extension/udfs/sequence_classification_single_text_udf.py
umitbuyuksahin: unique() does not reshuffle the data, so it preservers the order of the grouped by model_name.

exasol_transformers_extension/udfs/sequence_classification_single_text_udf.py
umitbuyuksahin: I thought about it and even tested the implementation on the same page. We may need to do a few workarounds, should I create a refactroring ticket without make this PR too complicated?

PredictionUDF is already decoupled from DownloaderUDF. So, two thread mechanism will be only useful for performing prediction on the GPU (as you said) and emitting results on CPU. We can investigate this in another ticket

tkilias (changes requested):

exasol_transformers_extension/udfs/sequence_classification_single_text_udf.py
```diff
@@ -54,24 +54,20 @@ def load_models(self, model_name, bucketfs_conn):
...
-        model_pred_df = pd.DataFrame(data=model_preds,
-                                     columns=['bucketfs_conn', 'model_name',
-                                              'text_data', 'label', 'score'])
-        return model_pred_df
+
+        tokens = self.last_loaded_tokenizer(
+            list(model_df['text_data']), return_tensors="pt")
+        logits = self.last_loaded_model(**tokens).logits
+        preds = torch.softmax(logits, dim=1).tolist()
+        labels_dict = self.last_loaded_model.config.id2label
```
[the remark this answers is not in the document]
umitbuyuksahin: yes, like that `{0: 'label_1', 1: 'label_2', 2: 'label_3', 3: 'label_4'}`

exasol_transformers_extension/udfs/sequence_classification_single_text_udf.py
```diff
@@ -54,24 +54,20 @@ def load_models(self, model_name, bucketfs_conn):
...
+        labels = list(map(lambda x: x[1], sorted(labels_dict.items())))
+
+        model_df = model_df.loc[
+            model_df.index.repeat(len(labels))].reset_index(drop=True)
+        model_df['label'] = labels * (model_df.shape[0]//len(labels))
+        model_df['score'] = sum(preds, [])
```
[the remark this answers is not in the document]
umitbuyuksahin: you understand correctly for first two ones
the last one is (a little trick) just to flatten the array from 2d to 1d. sum operation starts withe empty [] list and iteratively add items up.

I better to make here more readable

tests/unit_tests/udfs/test_sequence_classification_single_text.py
```diff
@@ -109,11 +89,17 @@ def test_sequence_classification_single_text_multiple_inputs(
...
     result = executor.run([Group(input_data)], exa)
 
-    n_labels = 2
+    n_labels = 4
     assert len(result[0].rows) == len(input_data) * n_labels \
-           and all(row[-1] == 0.5 for row in result[0].rows)
+           and result[0].rows[0] == (input_data[0] + ('label_1', 0.25)) \
+           and result[0].rows[1] == (input_data[0] + ('label_2', 0.25)) \
+           and result[0].rows[2] == (input_data[0] + ('label_3', 0.25)) \
+           and result[0].rows[3] == (input_data[0] + ('label_4', 0.25)) \
```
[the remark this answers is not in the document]
umitbuyuksahin: the all logits of these estimates are the same. softmax calculates the exp of all of them and then  and normalizes,  so they all naturally have same probability.

exasol_transformers_extension/udfs/sequence_classification_single_text_udf.py
```diff
@@ -29,54 +31,119 @@ def run(self, ctx):
...
+            -> pd.DataFrame:
+        """
+        Reformat the dataframe used in prediction, such that each input rows
+        has a row for each label and its probability score
+
+        :param model_df: Dataframe used in prediction
+        :param preds: List of prediction probabilities
+        :param labels: List of labels
+
+        :return: Prepared dataframe including input data and predictions
```
tkilias, line 133: Does it make sense to add an example input and output?

exasol_transformers_extension/udfs/sequence_classification_single_text_udf.py
```diff
@@ -0,0 +1,152 @@
...
+        """
+        n_labels = len(labels)
+
+        # Repeat each row consecutively as the number of labels. At the end,
+        # the dataframe is expanded from (m, n) to (m*n_labels, n)
+        repeated_indexes = model_df.index.repeat(n_labels)
+        model_df = model_df.loc[repeated_indexes].reset_index(drop=True)
+
+        # Fill the dataframe with labels repeatedly, such that each input rows
```
tkilias, line 142:
```suggestion
        # Fill the dataframe with labels repeatedly, such that each input row
```

exasol_transformers_extension/udfs/sequence_classification_single_text_udf.py
```diff
@@ -0,0 +1,152 @@
...
+        # has a row for each label
+        extension_factor = model_df.shape[0]//n_labels
```
tkilias, line 144: This should be the size of the df before the extension, if I am not mistaken. Maybe, store it in a variable input_row_count before the extension happens

exasol_transformers_extension/udfs/sequence_classification_single_text_udf.py
```diff
@@ -0,0 +1,152 @@
...
+        model_df['label'] = labels * extension_factor
+
+        # Flatten 2D prediction scores to 1D list and assign it to score
+        # column of the dataframe
+        preds_flatten = sum(preds, [])
+        model_df['score'] = [round(pred, 2) for pred in preds_flatten]
```
tkilias, line 150: don't round, if it is tedious for testing, then round before testing, but not in the udf

tests/unit_tests/udfs/test_sequence_classification_single_text.py
```diff
@@ -34,8 +34,7 @@ def from_pretrained(cls, model_name, cache_dir):
 
         @property
         def logits(self) -> torch.FloatTensor:
-            return torch.FloatTensor([
-                [0.1, 0.1, 0.1, 0.1], [0.1, 0.2, 0.3, 0.4]])
+            return torch.FloatTensor([[0.1, 0.2, 0.3, 0.4]])
```
tkilias, line 37: Now we only have one row per model. We need at best tests for many combinations, and these need to be different tests, probably the best strategy to implement that is via parameterized tests. One parameter, is likely the udf wrapper, because you need different mocks for the model

single model, single batch, batch complete
single model, single batch, batch incomplete
single model, multiple batch, last batch complete
single model, multiple batch, last batch incomplete
multiple model, single batch, last batch complete
multiple model, single batch, last batch incomplete
multiple model, multiple batch, last batch complete
multiple model, multiple batch, last batch incomplete
multiple model, multiple batch, multiple models per batch
multiple model, multiple batch, one model in multiple batches

tkilias (changes requested):

tests/unit_tests/udf_wrapper_params/sequence_classification/MockSequenceClassification.py
```diff
@@ -0,0 +1,36 @@
...
+            MockSequenceClassification.model_based_data_size_at_each_batch = \
+                kwargs['model_based_data_size_at_each_batch']
+
+    @classmethod
+    def from_pretrained(cls, model_name, cache_dir):
+        return cls
+
+    @property
+    def logits(self) -> torch.FloatTensor:
+        MockSequenceClassification.cnt += 1
```
tkilias, line 31: This is dangerous, can we do this differently?

tests/unit_tests/udf_wrapper_params/sequence_classification/MockSequenceClassification.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def __init__(self, **kwargs):
+        self.kwargs = kwargs
+        # initiated by tests
+        if 'model_based_data_size_at_each_batch' in kwargs:
+            MockSequenceClassification.cnt = 0
+            MockSequenceClassification.model_based_data_size_at_each_batch = \
+                kwargs['model_based_data_size_at_each_batch']
+
+    @classmethod
```
tkilias, line 25: I think, we don't need a class method here, we can give the SequenceClassificationSingleText also a factory class instead. And, then we probably don't need the cnt class property, anymore.

tests/unit_tests/udf_wrapper_params/sequence_classification/SingleModelMultipleBatchComplete.py
```diff
@@ -1,37 +1,55 @@
...
+            config=config,
+            logits=logits)
+    })
+
+    inputs = [("sub_dir1", "model1", "My test text")] * data_size
+
+    outputs = [("sub_dir1", "model1", "My test text", "label1", 0.21),
+               ("sub_dir1", "model1", "My test text", "label2", 0.24),
+               ("sub_dir1", "model1", "My test text", "label3", 0.26),
+               ("sub_dir1", "model1", "My test text", "label4", 0.29)] * 4
```
tkilias, line 53: Data_size

tests/unit_tests/udf_wrapper_params/sequence_classification/SingleModelMultipleBatchIncomplete.py
```diff
@@ -1,37 +1,56 @@
...
+            config=config,
+            logits=logits)
+    })
+
+    inputs = [("sub_dir1", "model1", "My test text")] * data_size
+
+    outputs = [("sub_dir1", "model1", "My test text", "label1", 0.21),
+               ("sub_dir1", "model1", "My test text", "label2", 0.24),
+               ("sub_dir1", "model1", "My test text", "label3", 0.26),
+               ("sub_dir1", "model1", "My test text", "label4", 0.29)] * 5
```
tkilias, line 53: Data_size

tests/unit_tests/udfs/test_sequence_classification_single_text.py
```diff
@@ -53,37 +53,17 @@ def create_mock_metadata(udf_wrapper):
...
+    MultipleModelMultipleBatchComplete,
+    MultipleModelMultipleBatchIncomplete,
+    MultipleModelSingleBatchComplete,
+    MultipleModelSingleBatchIncomplete,
+    MultipleModelMultipleBatchMultipleModelsPerBatch
 ])
-def test_sequence_classification_single_text_multiple_models_with_factory(
+def test_sequence_classification_single_text(
         params, upload_dummy_model_to_local_bucketfs):
     models_metadata = upload_dummy_model_to_local_bucketfs
```
tkilias, line 68: upload_dummy_model_to_local_bucketfs created files which we don't need for this test, please have a look at it, if we can simplify this

tkilias (changes requested):

tests/unit_tests/udfs/test_sequence_classification_single_text.py
umitbuyuksahin: we need a bucketfs_location in UDF. That's why we have to create a valid localfs in tests. But I updated it by removing redundant part in the fixture

tests/fixtures/model_fixture.py
```diff
@@ -38,24 +38,22 @@ def upload_model_to_local_bucketfs(download_model) -> PurePosixPath:
...
-
-            models_metadata.append((sub_dir, str(model_path)))
-        yield models_metadata
+        model_path = PurePosixPath(
+            tmpdir_name,
+            bucketfs_operations.get_model_path(sub_dir, model_params.name))
+        bucketfs_location = LocalFSMockBucketFSLocation(model_path)
+
+        for file_name, content in model_file_data_map.items():
+            bucketfs_location.upload_string_to_bucketfs(
```
tkilias, line 53: In this test we mock the model load completely, so we don't need a localbucketfs mock. Not sure if we use this function somewhere else.

tkilias: Or, do I miss something?

tkilias (changes requested):
Do we already have a with_db integration Test, I didn't find it.

tests/fixtures/model_fixture.py
umitbuyuksahin: we need to provide a valid bucketfs connection for location in here:

https://github.com/exasol/transformers-extension/blob/24e653f2efd980d17cc66e4dfeb89a11576fb1c4/exasol_transformers_extension/udfs/sequence_classification_single_text_udf.py#L68

tkilias: Ah, you are right. Do we need files for that? Or, only a valid URI?

umitbuyuksahin: valid URI, we can remove file uploading part

tkilias (changes requested):

tkilias (approved):

PR #13, opened 2022-07-20
#12: Update method for generating bucket udf path

### LINKED ISSUE #12: Update method for generating bucket udf path
We currently use bucketfs_utils to generate udf path
We should replace with the ` generate_udf_path` method
- this method will be added in this ticket https://github.com/exasol/bucketfs-utils-python/issues/72

umitbuyuksahin:
Fixes #12

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -12,12 +12,11 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-      - name: A workaround to increase available disk size
-        run: |
-          sudo rm -rf /usr/share/dotnet
-          sudo rm -rf /opt/ghc
-          sudo rm -rf "/usr/local/share/boost"
-          sudo rm -rf "$AGENT_TOOLSDIRECTORY"
+
+      - name: Free disk space
+        uses: jlumbroso/free-disk-space@main
+        with:
+          tool-cache: false
 
       - uses: actions/setup-python@v2
         with:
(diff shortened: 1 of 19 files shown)


.github/workflows/build.yaml
```diff
@@ -12,12 +12,11 @@ jobs:
...
     steps:
       - uses: actions/checkout@v2
-      - name: A workaround to increase available disk size
-        run: |
-          sudo rm -rf /usr/share/dotnet
-          sudo rm -rf /opt/ghc
-          sudo rm -rf "/usr/local/share/boost"
-          sudo rm -rf "$AGENT_TOOLSDIRECTORY"
+
+      - name: Free disk space
```
tkilias, line 16: Did we run into disk size issues , again?

tests/integration_tests/with_db/test_sequence_classification_single_text_script.py
```diff
@@ -12,7 +12,7 @@ def test_sequence_classification_single_text_script(
     for i in range(n_rows):
         input_data.append((
             bucketfs_conn_name,
-            model_params.sub_dir,
+            f"{model_params.sub_dir}",
```
tkilias, line 15: Using str() cast for converting to string, seems in this situation safer, because it is to add a other characters to the fstring

tkilias (changes requested):

.github/workflows/build.yaml
umitbuyuksahin: Yes, we had. 
We got the HTTPError: 400, just like the first time we got it.

PR #23, opened 2022-08-02
#18: Corrected model filtering in prediction UDFs

### LINKED ISSUE #18: Correct model filtering in prediction UDFs
- Currently we filter the given data based on only model_name in all prediction UDFs
- But different versions of each model can be stored under different bucketfs_conn connections and sub_dirs.
- Therefore model filtering should be extended by considering these details
- Extend tests accordingly.

Ex: filtering like that
```
unique_values = batch_df[['model','bucketfs_conn', 'sub_dir]].drop_duplicates().values
for model, bucketfs_conn, subdir in unique_values:
  model_df = batch_df[(batch_df['model']==model) & (..) ..]
```

umitbuyuksahin:
Fixes #18


tests/unit_tests/udf_wrapper_params/question_answering/multiple_model_locations.py
```diff
@@ -0,0 +1,54 @@
...
+
+class MultipleModelLocations:
+    """
+    multiple model, multiple batch, multiple models per batch
+    """
+    batch_size = 2
+    data_size = 1
+
+    input_data = \
+        [(None, "sub_dir1", "model1", "question", "context")] * data_size + \
```
tkilias, line 34: hm, unfortunately, we are currently injecting the bucketfs_conf in the test itself, maybe we should change that, because we aren't using it the bucketfs_con itself in this test and would need it as input here.

tkilias (changes requested):

tests/unit_tests/udf_wrapper_params/question_answering/multiple_locations_multiple_batch_complete.py
```diff
@@ -0,0 +1,59 @@
...
+        batch_size=params.batch_size,
+        pipeline=params.mock_pipeline,
+        base_model=params.mock_factory,
+        tokenizer=MockSequenceTokenizer)
+
+    def run(ctx: UDFContext):
+        udf.run(ctx)
+
+
+class MultipleModelLocationsMultipleBatchComplete:
```
tkilias, line 28: I think, we don't need tests for all combinations for batches and multiple models, I think, there was a misunderstanding, if we have them we can keep them.

tests/unit_tests/udf_wrapper_params/question_answering/multiple_locations_multiple_batch_complete.py
```diff
@@ -0,0 +1,59 @@
...
+    output_data = [("bfs_conn1", "sub_dir1", "model1", "question",
+                    "context", "answer 1", 0.1)] * data_size + \
+                  [("bfs_conn2", "sub_dir1", "model1", "question",
+                    "context", "answer 1", 0.1)] * data_size + \
+                  [("bfs_conn1", "sub_dir2", "model1", "question",
+                    "context", "answer 1", 0.1)] * data_size + \
+                  [("bfs_conn1", "sub_dir1", "model2", "question",
+                    "context", "answer 2", 0.2)] * data_size
+
+    mock_factory = MockQuestionAnsweringFactory({
```
tkilias, line 52: I think, we need a different factory here, because we need to check, if it uses the correct model accordingly the bucketfs_con, sub_dir and model. So, we need to return a different model (with different answer and score) for each combination of bucketfs_con, sub_dir and model

Applies also to all other MultiLocationTests

tkilias (changes requested):

tests/unit_tests/udf_wrapper_params/question_answering/multiple_locations_multiple_batch_complete.py
umitbuyuksahin: In the current situation, we have three different setup:
1. single model
2. multiple model
3. multiple locations (sub_dir, bfs combinations)

We should keep the third one, since we slice data based on a combination
I think, the second one can be merged with the third one.

tests/unit_tests/udf_wrapper_params/question_answering/multiple_locations_multiple_batch_complete.py
umitbuyuksahin: We can do it (different model for each combination) by keeping the current Factory class as it is. We just need to udpate the inputs. e.g.: 
```
    mock_factory = MockQuestionAnsweringFactory({
        "model1.1": MockQuestionAnsweringModel(answer="answer 1", score=0.1),
        "model1.2": MockQuestionAnsweringModel(answer="answer 2", score=0.2),
        ..
    })
```

tests/unit_tests/udf_wrapper_params/question_answering/multiple_locations_multiple_batch_complete.py
tkilias: I would like to keep the multiple models ones, because they are less functionality. As I said, I would add only the following multiple location test, to test exactly this feature. 

1.  MultipleBucketFSConSingleSubdirSingleModelNameSingleBatch : We switch the location in a single batch
2.  MultipleBucketFSConSingleSubdirSingleModelNameMultipleBatch : We switch the location in the second batch
3.  MultipleBucketFSConSingleSubdirSingleModelNameSingleBatch : We switch the location in a single batch
4.  SingleBucketFSConMultipleSubdirSingleModelNameMultipleBatch : We switch the location in the second batch

tests/unit_tests/udf_wrapper_params/question_answering/multiple_locations_multiple_batch_complete.py
tkilias: The important part to test is, same model name, but different location. See, https://github.com/exasol/transformers-extension/pull/23#discussion_r936504276

tests/unit_tests/udf_wrapper_params/question_answering/multiple_locations_multiple_batch_complete.py
umitbuyuksahin: we filter the data based on this function:
 `   
model_df = batch_df[
                (batch_df['model_name'] == model_name) &
                (batch_df['bucketfs_conn'] == bucketfs_conn) &
                (batch_df['sub_dir'] == sub_dir)]`
Is it necessary to have separated tests for each item of this combination to ensure that this filtering is working correctly?

I suggested combining MultipleLocations and MultipleModels tests, as I thought it would be sufficient to test different combinations in different batches.

tkilias: Tests should not assume any specific implementation and they should test as good as possible orthogonal features. Here, are the orthogonal features "handling multiple models with different batching" and "handling different locations"

tkilias (changes requested):
Does the following actually keep the order? Because, if not, we might load a model multiple times, because we don't run over the rows in the correct order. Maybe, the test factory should count, how often a model got loaded, such that we can check that it got only loaded ones.

```
unique_values = batch_df[
            ['model_name', 'bucketfs_conn', 'sub_dir']].drop_duplicates().values
```

tests/unit_tests/udf_wrapper_params/question_answering/multiple_bfsconn_single_subdir_single_model_multiple_batch.py
```diff
@@ -0,0 +1,64 @@
...
+    input_data = [(None, "bfs_conn1", "sub_dir1", "model1",
+                   "question", "context")] * data_size + \
+                 [(None, "bfs_conn2", "sub_dir1", "model1",
+                   "question", "context")] * data_size
+    output_data = [("bfs_conn1", "sub_dir1", "model1",
+                    "question", "context", "answer 1", 0.1)] * data_size + \
+                  [("bfs_conn2", "sub_dir1", "model1",
+                    "question", "context", "answer 2", 0.2)] * data_size
+
+    with tempfile.TemporaryDirectory() as tmpdir_name:
```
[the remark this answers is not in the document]
umitbuyuksahin: yes, it stay alive along the process but it just a tmp folder. a dummy path can even work

umitbuyuksahin:
yes it keeps order and simply extracts the combinations of the filtering.

model counting assert could be added to improve tests more

tests/unit_tests/udf_wrapper_params/question_answering/multiple_bfsconn_single_subdir_single_model_multiple_batch.py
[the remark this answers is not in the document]
tkilias: better use a dummy path, because that way you know which value to expect and don't confuse readers of the code

tkilias:
> yes it keeps order and simply extracts the combinations of the filtering.
> 
> model counting assert could be added to improve tests more

I suggest adding a comment to the line with a reference to the documentation of the function, which says it keeps the order. And, for the counters, I would they we create a extra ticket, because this PR is already huge.

exasol_transformers_extension/utils/dataframe_operations.py
```diff
@@ -0,0 +1,16 @@
...
+def get_sorted_unique_values(
+        df: pd.DataFrame, columns: List[T]) -> List[List[T]]:
+    """
+    Filter given dataframe and return the sorted unique values
+
+    :param df: Dataframe from which the uniques values are extracted
+    :param columns: List of columns to be filtered and sorted accordingly
+    """
+
+    return df[columns].drop_duplicates().sort_values(by=columns).values.tolist()
```
tkilias, line 16: I think, in the order by and here we should explicit define if we are ascending or descending sorting, because they both need to be the same

exasol_transformers_extension/udfs/sequence_classification_text_pair_udf.py
```diff
@@ -47,13 +50,22 @@ def get_batched_predictions(self, batch_df: pd.DataFrame) -> pd.DataFrame:
         :return: Prediction results of the corresponding dataframe
         """
         result_df_list = []
-        for model_name in batch_df['model_name'].unique():
-            model_df = batch_df[batch_df['model_name'] == model_name]
-
-            if self.last_loaded_model_name != model_name:
+        ordered_columns = constants.ORDERED_COLUMNS.split(",")
+        ordered_columns = constants.ORDERED_COLUMNS.split(",")
```
tkilias, line 54: duplicated line

exasol_transformers_extension/deployment/constants.py
```diff
@@ -16,4 +16,4 @@
         "question_answering_udf.jinja.sql",
 }
 
-
+ORDERED_COLUMNS = "model_name,bucketfs_conn,sub_dir"
```
tkilias, line 19: you can also use a list here, because jinja can work with dicts, lists and combinations of it

tkilias (changes requested):

exasol_transformers_extension/utils/dataframe_operations.py
umitbuyuksahin: Isn't both ascendening by default?

umitbuyuksahin: anyway, i would explicitly specify it as ascending

tests/unit_tests/utils/dataframe_operations.py
```diff
@@ -0,0 +1,31 @@
...
+sample_df = pd.DataFrame({
+    'A': sample_arr,
+    'B': sorted(sample_arr, reverse=True),
+    'C': list(range(len(sample_arr))),
+    'D': sample_arr
+})
+
+
+@pytest.mark.parametrize("columns, expected", [
+    (['A'], [[0], [1], [2], [3], [4], [5]]),
```
tkilias, line 16: it is helpful, if you can give your parameterization name, I usually, add a dummy str input , which describes what each single parameterization tests

tkilias (changes requested):

tests/unit_tests/utils/dataframe_operations.py
```diff
@@ -0,0 +1,36 @@
...
+        [[i] for i in range(len(sample_arr))]),
+    ("two_columns_different_sorting", ['A', 'B'],
+        [[0, 5],
+        [1, 5],
+        [2, 5],
+        [3, 4], [3, 5],
+        [4, 3], [4, 4],
+        [5, 0], [5, 1], [5, 2], [5, 3]]),
+])
+def test_get_sorted_unique_values(columns, expected):
```
tkilias:
```suggestion
def test_get_sorted_unique_values(description:str, columns:List[str], expected:List[List[int]]):
```

without the description parameter, the test should fail

tkilias (changes requested):

tkilias (approved):

Issue #19, opened 2022-07-29
Inherent prediction UDF classes from same base class


umitbuyuksahin:
# Background

- Prediction UDFs classes share many methods
- We should extract the common methods into an abstract class
- Each UDF class should extend this abstract class
- Note that this ticket is should be  dealt with once the implementation of all prediction classes is complete

# Acceptance Criteria
[ ] Create abstract class
[ ] Extend UDFs from the abstract class
[ ] Split large methods into small functions 
[ ] Rename UDF classes with _UDF_ suffix  
[ ] Remove warning in the shared method
[ ] Split get_batched_predictions into functions

PR #36, opened 2022-08-22
#33: Created translation udf

### LINKED ISSUE #33: Create Translation UDF
# Background
- Translation is the task of translating a text from one language to another.
- Transformers API provides [AutoModelForSeq2SeqLM](https://huggingface.co/docs/transformers/task_summary#translation) class to perform translation

# Acceptance Criteria
- Create UDF for Translation
- Support both multilingual and single-pair models
- Add unit/integration tests

umitbuyuksahin:
Fixes #33


tests/integration_tests/with_db/udfs/test_translation_script.py
```diff
@@ -0,0 +1,51 @@
...
+def test_translation_script(
+        upload_language_container, setup_database,
+        pyexasol_connection, upload_model_to_bucketfs):
+
+    bucketfs_conn_name, schema_name = setup_database
+    n_rows = 100
+    model_name = "t5-small"
+    src_lang = "English"
+    tgt_lang = "German"
+    max_length = 50
```
tkilias, line 16: why do need max length for translation. Exists this parameter, really.

tkilias (commented):
Added a question

tests/integration_tests/with_db/udfs/test_translation_script.py
umitbuyuksahin: It is used in almost all translation examples. If the length is short, it summarizes the given text accordingly. I thought it might be useful.

tests/integration_tests/without_db/udfs/test_sequence_classification_single_text_udf.py
```diff
@@ -41,7 +41,11 @@ def get_dataframe(self, num_rows='all', start_col=0):
         return return_df
 
 
-@pytest.mark.parametrize("device_id", [None, 0])
+@pytest.mark.parametrize(
+    "device_id, upload_model_to_local_bucketfs", [
+        (None, model_params.base),
+        (0, model_params.base)
+    ], indirect=["upload_model_to_local_bucketfs"])
```
tkilias, line 48: and using two different fixtures wasn't and option, it probably would have easier and would look less confusing

exasol_transformers_extension/udfs/models/translation_udf.py
```diff
@@ -0,0 +1,188 @@
...
+        self.pipeline = pipeline
+        self.base_model = base_model
+        self.tokenizer = tokenizer
+        self.device = None
+        self.cache_dir = None
+        self.last_loaded_model_key = None
+        self.last_loaded_model = None
+        self.last_loaded_tokenizer = None
+        self.last_created_pipeline = None
+        self.translation_prefix = "translate {src_lang} to {tgt_lang}: "
```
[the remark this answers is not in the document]
umitbuyuksahin: Yeah. According to multi-lang models, it may be necessary to call API  differently. For example:

While it is necessary to give src_lang/tgt_lang parameters in the __call__ method of the m2m100 model, these parameters do not work in the t5 model. 

I've organized it according to the most common usage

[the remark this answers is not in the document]
tkilias: ok, that is fine for now, but maybe we can determine the common ways and add a parameter to switch the behavior, but I would move this to a ticket. Further, we need to document this assumption to the user, by stating which models are compatible and which method for encoding the language we are using

tests/integration_tests/without_db/udfs/test_translation_udf.py
```diff
@@ -0,0 +1,111 @@
...
+    bucketfs_base_path = upload_model_seq2seq_to_local_bucketfs
+    bucketfs_conn_name = "bucketfs_connection"
+    bucketfs_connection = Connection(address=f"file://{bucketfs_base_path}")
+
+    batch_size = 2
+    sample_data = [(
+        None,
+        bucketfs_conn_name,
+        model_params.sub_dir,
+        "t5-small",
```
tkilias, line 81: Didn't you get this from model_params in the other tests?

tests/unit_tests/udf_wrapper_params/translation/mock_translation.py
```diff
@@ -0,0 +1,55 @@
...
+            "German:": "übersetzt",
+            "French:": "traduit"
+        }
+
+    def __call__(self, text_data: List[str], **kwargs) -> List[Dict[str, str]]:
+        max_len = kwargs["max_length"]
+        results = []
+        for text in text_data:
+            splitted_text = text.split()
+            tgt_lang = splitted_text[3]
```
tkilias, line 50:
```suggestion
            target_lang = splitted_text[3]
```

tests/unit_tests/udf_wrapper_params/translation/multiple_bfsconn_single_subdir_single_model_multiple_batch.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+class MultipleBucketFSConnSingleSubdirSingleModelNameMultipleBatch:
+    """
+    multiple bucketfs connection, single subdir, single model, multiple_batch
+    """
+    batch_size = 2
+    data_size = 2
+    src_lang = "English"
+    tgt_lang = "German"
```
tkilias, line 35:
```suggestion
    target_lang = "German"
```

tkilias: Please change all other places where you used this, too

tests/utils/parameters.py
```diff
@@ -31,7 +31,8 @@ def address(self, port=None) -> str:
 
 @dataclass(frozen=True)
 class ModelParams:
-    name: str
+    base: str
+    seq2seq: str
```
tkilias, line 35:
```suggestion
    seq2seq_model: str
```

tkilias (changes requested):

exasol_transformers_extension/udfs/models/translation_udf.py
```diff
@@ -0,0 +1,188 @@
...
+            current_model_key = (bucketfs_conn, sub_dir, model_name)
+            if self.last_loaded_model_key != current_model_key:
+                self.set_cache_dir(model_df)
+                self.clear_device_memory()
+                self.load_models(model_name)
+                self.last_loaded_model_key = current_model_key
+
+            unique_params = dataframe_operations.get_unique_values(
+                model_df, ['max_length', 'source_language', 'target_language'])
+            for max_length, source_language, target_language in unique_params:
```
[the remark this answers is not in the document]
umitbuyuksahin: will be handled in refactoring ticket #19

tests/fixtures/model_fixture.py
```diff
@@ -1,64 +1,105 @@
...
     LocalFSMockBucketFSLocation
-from exasol_transformers_extension.utils import bucketfs_operations
-from tests.utils.parameters import model_params
+from exasol_bucketfs_utils_python.abstract_bucketfs_location import \
+    AbstractBucketFSLocation
 
 
 @pytest.fixture(scope="session")
-def download_model() -> str:
+def get_local_bucketfs_path() -> str:
```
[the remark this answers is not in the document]
umitbuyuksahin: it is leftover, removed

tests/fixtures/model_fixture.py
```diff
@@ -1,64 +1,105 @@
...
+def upload_model_base_to_bucketfs(
+        bucketfs_location: AbstractBucketFSLocation) -> PurePosixPath:
 
-        yield str(model_path)
+    with upload_model_to_bucketfs(
+            model_params.base, bucketfs_location) as path:
+        try:
+            yield path
+        finally:
+            _cleanup_buckets(bucketfs_location, path)
```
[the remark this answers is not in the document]
umitbuyuksahin: wouldn't it get the cleanup run right after yield if I move to. In the current implementation, it is kept until end of session then it runs.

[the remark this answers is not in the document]
tkilias: so, you would need  to move the `try finally`, as well. And, the yield in `upload_model_to_bucketfs` only returns when the `with upload_model_to_bucketfs` block finishes

tests/unit_tests/udf_wrapper_params/translation/single_model_single_batch_complete.py
```diff
@@ -0,0 +1,55 @@
...
+    def run(ctx: UDFContext):
+        udf.run(ctx)
+
+
+class SingleModelSingleBatchComplete:
+    """
+    single model, single batch, batch complete
+    """
+    batch_size = 2
+    data_size = 1
```
tkilias, line 33: I think, this is wrong, SingleModel, SingleBatch, Complete should be 

```suggestion
    batch_size = 2
    data_size = 2
```

tkilias (changes requested):

tests/fixtures/model_fixture.py
```diff
@@ -3,6 +3,9 @@
 import transformers
 from contextlib import contextmanager
 from pathlib import PurePosixPath, Path
+
+from py._path.local import LocalPath
```
tkilias, line 7: what is that? looks like a private module

tests/fixtures/model_fixture.py
```diff
@@ -41,58 +33,63 @@ def upload_model(bucketfs_location: AbstractBucketFSLocation,
...
 
 @contextmanager
-def upload_model_to_local_bucketfs(model_name: str) -> str:
-    with download_model(model_name) as download_tmpdir:
-        with tempfile.TemporaryDirectory() as upload_tmpdir_name:
-            bucketfs_location = LocalFSMockBucketFSLocation(upload_tmpdir_name)
-            upload_model(bucketfs_location, model_name, download_tmpdir)
-            yield upload_tmpdir_name
+def upload_model_to_local_bucketfs(
+        model_name: str, download_tmpdir: LocalPath) -> str:
```
tkilias, line 37: use PosixPurePath or Path instead of LocalPath

tests/fixtures/model_fixture.py
```diff
@@ -41,58 +33,63 @@ def upload_model(bucketfs_location: AbstractBucketFSLocation,
...
+
+    download_model(model_name, download_tmpdir)
+    with tempfile.TemporaryDirectory() as upload_tmpdir_name:
```
tkilias, line 40: you could use a subdir of download_tmpdir, instead of creating a new directory

pyproject.toml
```diff
@@ -34,5 +34,5 @@ exasol-udf-mock-python = { git = "https://github.com/exasol/udf-mock-python.git"
 exasol-script-languages-container-tool = { git = "https://github.com/exasol/script-languages-container-tool.git", branch = "0.14.0"}
 
 [build-system]
-requires = ["poetry-core>=1.0.0"]
+requires = ["poetry-core>=1.2.0"]
```
tkilias: this needs to stay on 1.0.0, poetry-core is different from the poetry cli

tkilias (changes requested):

tkilias (approved):

PR #41, opened 2022-09-15
#19: Inherented prediction udf classes from same base class

### LINKED ISSUE #19: Inherent prediction UDF classes from same base class

umitbuyuksahin:
Fixes #19

### DIFF
--- a/exasol_transformers_extension/udfs/models/base_model_udf.py
+++ b/exasol_transformers_extension/udfs/models/base_model_udf.py
@@ -1,4 +1,4 @@
-from abc import abstractmethod
+from abc import abstractmethod, ABC
 from typing import Iterator, List, Any, Optional
 import torch
 import pandas as pd
@@ -7,7 +7,7 @@ from exasol_transformers_extension.utils import device_management, \
     bucketfs_operations, dataframe_operations
 
 
-class BaseModelUDF:
+class BaseModelUDF(ABC):
     def __init__(self,
                  exa,
                  batch_size,
@@ -85,35 +85,23 @@ class BaseModelUDF:
 
             current_model_key = (bucketfs_conn, sub_dir, model_name)
             if self.last_loaded_model_key != current_model_key:
-                self.set_cache_dir(model_df)
+                self.set_cache_dir(model_name, bucketfs_conn, sub_dir)
                 self.clear_device_memory()
                 self.load_models(model_name)
                 self.last_loaded_model_key = current_model_key
 
             yield model_df
 
-    def extract_unique_param_based_dataframes(
-            self, model_df: pd.DataFrame) -> Iterator[pd.DataFrame]:
-        """
-        Extract unique dataframes having same model parameter values. if there
-        is no model specified parameter, the input dataframe return as it is.
-
-        :param model_df: Dataframe used in prediction
-
-        :return: Unique model dataframes having specified parameters
-        """
-
-        yield model_df
-
-    def set_cache_dir(self, model_df: pd.DataFrame) -> None:
+    def set_cache_dir(
+            self, model_name: str, bucketfs_conn_name: str,
+            sub_dir: str) -> None:
         """
         Set the cache directory in bucketfs of the specified model.
 
-        :param model_df: The model dataframe to set the cache directory
+        :param model_name: Name of the model to be cached
+        :param bucketfs_conn_name: Name of the bucketFS connection
+        :param sub_dir: Directory where the model is cached
         """
-        model_name = model_df['model_name'].iloc[0]
-        bucketfs_conn_name = model_df['bucketfs_conn'].iloc[0]
-        sub_dir = model_df['sub_dir'].iloc[0]
         bucketfs_location = bucketfs_operations.create_bucketfs_location(
             self.exa.get_connection(bucketfs_conn_name))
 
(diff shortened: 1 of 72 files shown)


exasol_transformers_extension/udfs/models/base_model_udf.py
```diff
@@ -0,0 +1,196 @@
...
+                result_df_list.append(result_df)
+
+        result_df = pd.concat(result_df_list)
+        return result_df
+
+    def extract_unique_model_dataframes_from_batch(
+            self, batch_df: pd.DataFrame) -> Iterator[pd.DataFrame]:
+        """
+        Extract unique model dataframes with the same model_name, bucketfs_conn,
+        and sub_dir from the dataframe. If the extracted model is not cached,
```
tkilias, line 69: The cache dir likely points to /buckets/... which is read only so a download of a non cached  model likely fails.

tkilias, line 69: The cache dir likely points to /buckets/... which is read only so a download of a non cached  model likely fails.

exasol_transformers_extension/udfs/models/base_model_udf.py
```diff
@@ -0,0 +1,196 @@
...
+            current_model_key = (bucketfs_conn, sub_dir, model_name)
+            if self.last_loaded_model_key != current_model_key:
+                self.set_cache_dir(model_df)
+                self.clear_device_memory()
+                self.load_models(model_name)
+                self.last_loaded_model_key = current_model_key
+
+            yield model_df
+
+    def extract_unique_param_based_dataframes(
```
tkilias, line 95: Should be this method abstract, because tbe default implementation is not complex and with it , new implementation might not know about it

exasol_transformers_extension/udfs/models/base_model_udf.py
```diff
@@ -0,0 +1,196 @@
...
+        is no model specified parameter, the input dataframe return as it is.
+
+        :param model_df: Dataframe used in prediction
+
+        :return: Unique model dataframes having specified parameters
+        """
+
+        yield model_df
+
+    def set_cache_dir(self, model_df: pd.DataFrame) -> None:
```
tkilias, line 108: I think the modeldf parameters is overkill, this method only needs model_name, bucketfs_conn, sub_dir. If you only give this to the function it also more clear what it needs

exasol_transformers_extension/udfs/models/base_model_udf.py
```diff
@@ -0,0 +1,196 @@
...
+    def clear_device_memory(self):
+        """
+        Delete models and free device memory
+        """
+
+        del self.last_loaded_model
+        del self.last_loaded_tokenizer
+        torch.cuda.empty_cache()
+
+    def load_models(self, model_name: str, **kwargs) -> None:
```
tkilias, line 133: You have defined the kwargs but don't use them, is there are reason

exasol_transformers_extension/udfs/models/base_model_udf.py
```diff
@@ -0,0 +1,196 @@
...
+        :return: The dataframe where the model_df is formatted with the
+        prediction results
+        """
+        pred_df_list = self.execute_prediction(model_df)
+        pred_df = self.append_predictions_to_input_dataframe(
+            model_df, pred_df_list)
+        return pred_df
+
+    @staticmethod
+    def create_dataframes_from_predictions(
```
tkilias, line 166: Looks like this function is not used in the base class, if it is used in the subclass extract it as a utility function

exasol_transformers_extension/udfs/models/base_model_udf.py
```diff
@@ -0,0 +1,196 @@
+from abc import abstractmethod
+from typing import Iterator, List, Any, Optional
+import torch
+import pandas as pd
+from exasol_transformers_extension.deployment import constants
+from exasol_transformers_extension.utils import device_management, \
+    bucketfs_operations, dataframe_operations
+
+
+class BaseModelUDF:
```
tkilias, line 10: Needs to inherit from ABC

tkilias (changes requested):
I had a deeper look at the BaseModelUDF class. The implementation it self, I couldn't check in depth because the diff of GitHub is not readable. I need to check this on Monday in the IDE.

exasol_transformers_extension/udfs/models/base_model_udf.py
umitbuyuksahin: agree, thx

exasol_transformers_extension/udfs/models/base_model_udf.py
umitbuyuksahin: I didn't face this problem. cache files are located under /cotainer folder.

tkilias: but probably not inside the UDFs.

exasol_transformers_extension/udfs/models/base_model_udf.py
tkilias: but, if it is not used in the abstract class, than we don't need to define it as abstract.

umitbuyuksahin: It is necessary to create dataframes according to each model result. Is it better to put each one separately into utility or leave it in its own class?

tkilias (approved):

PR #49, opened 2022-09-22
#48: Prepared the initial release

### LINKED ISSUE #48: Prepare the first release
- add build configuration scripts
- complete change log

umitbuyuksahin:
Fixes #48

### DIFF
--- a/doc/changes/changes_0.1.0.md
+++ b/doc/changes/changes_0.1.0.md
@@ -4,12 +4,20 @@ Code name: Add a downloader UDF and a set of prediction UDFs using the transform
 
 
 ## Summary
+
 This is the initial release of the transformers-extension which provides a 
 downloader UDF that allows us to cache the pre-trained machine learning model 
 by transformers and a set of Prediction UDFs that allow the use of the cached 
-machine learning models via the transformers api. This version provides the 
-following machine learning tasks: Sequence Classification, Question Answering, 
-Filling Mask, Text Generation, Token Classification, and Text Translation.
+machine learning models via the transformers API.
+
+This version provides the following machine learning tasks:
+
+* Sequence Classification
+* Question Answering, 
+* Filling Mask
+* Text Generation
+* Token Classification
+* Text Translation.
 
 ### Features
 
--- a/doc/changes/changes_0.1.0.md
+++ b/doc/changes/changes_0.1.0.md
@@ -1,6 +1,6 @@
-# transformers-extension 0.1.0, released 2022-09-22
+# Transformers Extension 0.1.0, released 2022-09-22
 
-Add a downloader UDF and a set of prediction UDFs using the transformers API
+Code name: Add a downloader UDF and a set of prediction UDFs using the transformers API
 
 
 ## Summary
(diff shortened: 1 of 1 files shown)


doc/changes/changes_0.1.0.md
```diff
@@ -1,8 +1,15 @@
-# transformers-extension 0.1.0, released XXXX-XX-XX
+# Transformers Extension 0.1.0, released 2022-09-23
 
+Code name: Add a downloader UDF and a set of prediction UDFs using the transformers API
 
-## Summary
 
+## Summary
+This is the initial release of the transformers-extension which provides a
```
redcatbear, line 7:
```suggestion
## Summary

This is the initial release of the transformers-extension which provides a 
```

The original Markdown spec requires a blank line between a headline and a text body. Most parsers are not that picky though. :-)

doc/changes/changes_0.1.0.md
```diff
@@ -1,8 +1,15 @@
-# transformers-extension 0.1.0, released XXXX-XX-XX
+# Transformers Extension 0.1.0, released 2022-09-23
 
+Code name: Add a downloader UDF and a set of prediction UDFs using the transformers API
 
-## Summary
 
+## Summary
+This is the initial release of the transformers-extension which provides a 
+downloader UDF that allows us to cache the pre-trained machine learning model
```
redcatbear, line 8: Is this really a cache? I see it more as pre-loading existing models. A cache would be in my eyes if the model was auto-generated once and then kept in a cache without user interaction.

doc/changes/changes_0.1.0.md
```diff
@@ -1,8 +1,15 @@
...
 
-## Summary
 
+## Summary
+This is the initial release of the transformers-extension which provides a 
+downloader UDF that allows us to cache the pre-trained machine learning model 
+by transformers and a set of Prediction UDFs that allow the use of the cached 
+machine learning models via the transformers api. This version provides the 
+following machine learning tasks: Sequence Classification, Question Answering, 
+Filling Mask, Text Generation, Token Classification, and Text Translation.
```
redcatbear, line 12:
```suggestion
machine learning models via the transformers API.

This version provides the following machine learning tasks:

* Sequence Classification
* Question Answering, 
* Filling Mask
* Text Generation
* Token Classification
* Text Translation.
```

redcatbear (changes requested):

doc/changes/changes_0.1.0.md
tkilias: Yes and no. It would be an cache, except that is not suitable for how storage works in the UDFs. And, on the other hand, we use the cache facilities of the transformers library. Usually, the cache would also include eviction, which is actually not part of the cache functionality of the transformers library. However, users if the libraries probably now about the cache functionality and are used to the term. 

Maybe, we should write something like preload the model to the cache.

redcatbear: :heavy_check_mark:

redcatbear (approved):

tkilias (approved):

PR #53, opened 2022-09-23
#52: Added GH actions that free disk space

### LINKED ISSUE #52: Reduce disk space used by the machine during releasing
- we need to free disk space of the machine while releasing.
- otherwise, SLC and other installed large-sized libraries cause of disk problem

umitbuyuksahin:
Fixes #52


morazow (approved):
:+1:

PR #74, opened 2023-02-01
#73: Added one step installation

### LINKED ISSUE #73: Deploy Language Container in One Step
# Background
- In the current language container installation, we firstly download artifacts from Github Repo. We then have to run deployment script.
- The deployment script can contain the downloading step, so that users are able to install the SLC in one step

# Acceptance Criteria
- Update language container script  to get artifact version and download
- Update user guide, enhance manual and script deployments

umitbuyuksahin:
Fixes #73


exasol_transformers_extension/deployment/utils.py
```diff
@@ -9,6 +15,10 @@
 
 DB_PASSWORD_ENVIRONMENT_VARIABLE = f"TE_DB_PASSWORD"
 BUCKETFS_PASSWORD_ENVIRONMENT_VARIABLE = f"TE_BUCKETFS_PASSWORD"
+SLC_PARTS_PREFIX_NAME = 'language_container_part_0'
+SLC_FINAL_NAME = "language_container.tar.gz"
+N_SLC_PARTS = 2
+GH_RELEASE_URL = "https://github.com/exasol/transformers-extension/releases/download"
```
Nicoretti, line 21: reading parameters also from `ENV` would enable the user(s) to have a `.env` file which can be put into `.gitingore` which would result in
* Less configuration within the code
* Less risk of leaking confidential settings (e.g. password)
* Less arguments to pass for the regular CLI invocations

exasol_transformers_extension/deployment/utils.py
```diff
@@ -28,3 +38,32 @@ def load_and_render_statement(template_name, **kwargs) -> str:
...
+
+
+def _download_slc_parts(tmp_dir, version):
+    for i in range(N_SLC_PARTS):
+        slc_part_name = f"{SLC_PARTS_PREFIX_NAME}" + str(i)
+        url = "/".join((GH_RELEASE_URL, version, slc_part_name))
+        response = requests.get(url, stream=True)
+        with open(Path(tmp_dir, slc_part_name), 'wb') as f:
+            f.write(response.content)
+
```
Nicoretti, line 50:
```suggestion
def _download_slc_parts(directory, version):
    parts = (f"{SLC_PARTS_PREFIX_NAME}{index}" for index in range(N_SLC_PARTS))
    urls = ("/".join((GH_RELEASE_URL, version, part)) for part in parts)
    for url in urls:
        response = requests.get(url, stream=True)
        response.raise_for_status()
        with open(Path(directory, slc_part_name), 'wb') as f:
            f.write(response.content)
```

exasol_transformers_extension/deployment/utils.py
```diff
@@ -28,3 +38,32 @@ def load_and_render_statement(template_name, **kwargs) -> str:
...
+def _concatenate_slc_parts(tmp_dir):
+    slc_final_path = Path(tmp_dir, SLC_FINAL_NAME)
+    slc_parts_path = glob.glob(str(Path(tmp_dir, SLC_PARTS_PREFIX_NAME)) + "*")
+    cmd = f"cat {' '.join(slc_parts_path)} > {slc_final_path}"
+    subprocess.run(cmd,
+                   stdout=subprocess.PIPE,
+                   stderr=subprocess.STDOUT,
+                   shell=True)
+    return slc_final_path
```
Nicoretti, line 60:
```suggestion
import fileinput

def _concatenate_slc_parts(directory):
    destination = Path(directory) / SLC_FINAL_NAME
    parts = glob.glob(f"{Path(directory) / SLC_PARTS_PREFIX_NAME}*")
    with fileinput.input(files=parts)) as input:
        with open(destination, 'w') as output:
            output.writelines(inputs)
    return destination
```

exasol_transformers_extension/deployment/utils.py
```diff
@@ -28,3 +38,32 @@ def load_and_render_statement(template_name, **kwargs) -> str:
...
+
+
+@contextmanager
+def get_container_file_from_github_release(version):
+    with tempfile.TemporaryDirectory() as tmp_dir:
+        _download_slc_parts(tmp_dir, version)
+        container_file_path = _concatenate_slc_parts(tmp_dir)
+        yield container_file_path
```
Nicoretti, line 68:
```suggestion
@contextmanager
def get_container_file_from_github_release(version):
    with tempfile.TemporaryDirectory() as tmp_dir:
        _download_slc_parts(tmp_dir, version)
        path = _concatenate_slc_parts(tmp_dir)
        yield path
```

tests/utils/revert_language_settings.py
```diff
@@ -4,16 +4,15 @@
...
+    def wrapper(**kwargs):
         try:
-            return func(language_alias, schema, db_conn,
-                        container_path, language_settings)
+            return func(**kwargs)
         except Exception as exc:
             print("Exception occurred while running the test: %s" % exc)
             raise pytest.fail(exc)
         finally:
             print("Revert language settings")
```
Nicoretti, line 14: generally I only would use prints in the main function, within library code I would suggest the use of logging.

umitbuyuksahin: thanks

exasol_transformers_extension/deployment/utils.py
umitbuyuksahin: loops are small but don't you think it would be more confusing to increase to 3 loops?

exasol_transformers_extension/deployment/language_container_deployer_cli.py
```diff
@@ -30,33 +31,46 @@ def language_container_deployer_main(
...
+    if container_file:
+        call_runner()
+    elif version:
+        with utils.get_container_file_from_github_release(version) as container:
+            container_file = container
+            call_runner()
+    else:
+        raise Exception("You should specify either the release version to "
+                        "download container file or the path of the already "
+                        "downloaded container file.")
```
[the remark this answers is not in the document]
umitbuyuksahin: From readability point of view, the first version seems better for me. Why did you suggest this?

exasol_transformers_extension/deployment/utils.py
umitbuyuksahin: good idea, but there many configurable constants. it is better to handle in different ticket

exasol_transformers_extension/deployment/language_container_deployer_cli.py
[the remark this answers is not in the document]
Nicoretti: Because a reader can think about the conditions one at the time and therefore from my point of view it is easier to understand when and what happens. Also it is clear when `call_runner()` is invoked. Also it is ensured that when `call_runner()` is invoked `container_file == True`, this is not guaranteed in the `elif` branch.

TL;DR: I think the individual parts are can be better reasoned about and we do not depend on an other function to ensure we don't crash.

exasol_transformers_extension/deployment/utils.py
Nicoretti: A generator is lazy so it is not actually a loop in that sense. I personally find it easier and better to read and I think it separates the individual concerns better. Still for sure also partially matter of taster. Performance wise I think neither version  will be much faster. If you are concerned regarding performance I would recommend to download the parts asynchronously.

:warning: If you want to keep it as is that is fine with me, but I suggest you at least add the **_status check_** `response.raise_for_status()` .

Nicoretti (approved):

PR #81, opened 2023-02-15
#72:  Added user token to download model from private repos

### LINKED ISSUE #72: Download models from private libraries
## Background 

We can download only open models (without any authentication) from the Huggingface repository. We should be able to download those models from private repos by passing authentication strings.

umitbuyuksahin:
Fixes #72


Nicoretti (dismissed):
A general remark: Having a code formatter (ideally black) and an import sorter (e.g. isort) continuously applied to the code base would be good.

Nicoretti (approved):

PR #92, opened 2023-04-20
#24: Added model counters to unit tets

### LINKED ISSUE #24: Add model counters to unit tests of prediction UDFs
- In the prediction UDFs, input dataframe is filtered based on _model_name, bucketfs_conn, sub_dir_ features
- This filtering is already implemented to prevent same model multiple times in each batch iteration
- To improve tests in this directions, tests cases should be enhanced with model counter assertion

for FillingMask 
- https://github.com/exasol/transformers-extension/blob/0df0d6e78f6c8656d10926653ff02612bbc416e9/tests/unit_tests/udf_wrapper_params/filling_mask/mock_filling_mask.py#L27

umitbuyuksahin:
Fixes #24

### DIFF
--- a/doc/changes/changes_0.5.0.md
+++ b/doc/changes/changes_0.5.0.md
@@ -14,6 +14,10 @@ TBD
 ### Bug Fixes
 
  - #89: Fixed the content of error code config file 
+
+### Refactorings
+
+ - #24: Added model counters to unit tests of prediction UDFs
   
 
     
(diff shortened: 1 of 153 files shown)


Nicoretti (approved):
Maybe some of the test could be restructured to be [parameterized tests](https://docs.pytest.org/en/6.2.x/parametrize.html) this could simplify changes like this.

umitbuyuksahin:
Different scenarios are already parameterized for each model test. The code duplication comes from having to have a separate udf_wrapper for each scenario.

PR #98, opened 2023-06-05
95: Removed setup.py

### LINKED ISSUE #95: Remove setup.py
# Background

- with poetry 1.4.0 it doesn't create the setup.py anymore https://github.com/python-poetry/poetry/releases/tag/1.4.0
- we currently use poetry build to generate the setup.py
- however, the setup.py isn't needed with newer pip versions and if there are releases to pypi or as wheels
- for that reason, we can remove setup.py and githook that generates it, from this repo

# Acceptance Criteria

- [x] Update workflows to poetry 1.4.0
- [x] Remove setup.py Github Workflow
- [x] Remove setup.py githook
- [x] Remove setup.py file

MarleneKress79789:
fixes #95


.github/workflows/build.yaml
```diff
@@ -32,7 +32,7 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - uses: abatilo/actions-poetry@v2.1.4
         with:
-          poetry-version: 1.2.2
+          poetry-version: 1.4.0
```
tkilias, line 35: Can you use the following action https://github.com/exasol/integration-test-docker-environment/blob/main/.github/actions/prepare_poetry_env/action.yml and copy it to this repo

tkilias: and use it also it this file 
https://github.com/exasol/transformers-extension/blob/62e126ed475123433eb6a23ca2f42857002625fc/.github/workflows/release_droid_upload_github_release_assets.yml

MarleneKress79789: done

tkilias (approved):

PR #102, opened 2023-07-21
#100: Enabled encryption for all pyexasol connection to be compatible with Exasol 8

### LINKED ISSUE #100: ScriptDeployer Client Connection needs to be encrypted
# Background

- When running the 

  ```
  python -m exasol_transformers_extension.deploy scripts --dsn 127.0.0.1:9564 --db-user sys \
  --db-pass exasol --schema TRASNFORMERS \
  --language-alias transformers
  ```
  
  the following error message appears with Docker-DB 8.18.1
  
  image (19)

  reported by @exa-eswar

- probable reason is too old a version of pyexasol or a misconfigured pyexasol connection

tkilias:
Fixes #100


exasol_transformers_extension/deployment/language_container_deployer.py
```diff
@@ -93,7 +94,14 @@ def run(cls, bucketfs_name: str, bucketfs_host: str, bucketfs_port: int,
...
         pyexasol_conn = pyexasol.connect(
-            dsn=dsn, user=db_user, password=db_password)
+            dsn=dsn,
+            user=db_user,
+            password=db_password,
+            encryption=True,
+            websocket_sslopt={
+                "cert_reqs": ssl.CERT_NONE,
+            }
+        )
```
Nicoretti, line 104: For every functionality which is not test related, I would recommend enable the user of the functionality to decide on enabling/disabling the certificate check.

exasol_transformers_extension/deployment/scripts_deployer.py
```diff
@@ -48,6 +49,14 @@ def deploy_scripts(self) -> None:
...
             schema: str, language_alias: str):
 
-        pyexasol_conn = pyexasol.connect(dsn=dsn, user=user, password=password)
+        pyexasol_conn = pyexasol.connect(
+            dsn=dsn,
+            user=user,
+            password=password,
+            encryption=True,
+            websocket_sslopt={
+                "cert_reqs": ssl.CERT_NONE,
```
Nicoretti, line 58: For every functionality which is not test related, I would recommend enable the user of the functionality to decide on enabling/disabling the certificate check.

Nicoretti (changes requested):

exasol_transformers_extension/deployment/language_container_deployer.py
tkilias: Agree, will do this on another ticket

tkilias: https://github.com/exasol/transformers-extension/issues/103

Nicoretti (approved):

PR #104, opened 2023-07-24
#42: Update transformers to 4.31 and adapt the model uploader

### LINKED ISSUE #42: Support Transformers >= 4.22
# Background

- The directory structure where  pre-trained models are downloaded has been changed with transformers 4.22
- While the models files directly are downloaded into the cache directory in older versions, with version 4.22 transformers uses complex directory structure.
- We need to upload this directory to bucket as we downloaded.
- In order to complete this issue,  the ticket https://github.com/exasol/bucketfs-python/issues/1 must be done.

# Acceptance Criteria

- switched to upload_directory method in bucketfs-utils
- update toml file with `transformers="^4.22"'
  -    pay attention that newer versions install nvidia modules which increases SLC size ~2GB

### LINKED ISSUE #82: Add integratoin tests for private models
# Background
- Current tests are working with public models
- To have tests for private models we need a valid user authentication token
- We can create a private test repo in huggingface hub so that we can perform integration tests for private models.
 
# Acceptance Criteria
- create a private model repo
- generate a token, put it into env so that we can get it from github secrets
- implement integration tests
https://huggingface.co/docs/huggingface_hub/how-to-manage

tkilias:
Fixes #42


tests/unit_tests/utils/test_bucketfs_operations.py
```diff
@@ -0,0 +1,69 @@
...
+            'test_model_path/test_model_name/blobs/234608c922aaf3989d6a772af31711fbbdd62e3a'),
+        call.upload_string_to_bucketfs(
+            'test_model_path/test_model_name/.no_exist/6f75de8b60a9f8a2fdf7b69cbd86d9e64bcb3837/.not_empty', ''),
+        call.upload_fileobj_to_bucketfs(
+            unittest.mock.ANY,
+            'test_model_path/test_model_name/.no_exist/6f75de8b60a9f8a2fdf7b69cbd86d9e64bcb3837/tokenizer_config.json')
+    ]
+
+
+def create_no_exits_directory(model_name, ref, tmp_path):
```
MarleneKress79789, line 45: typo?
```suggestion
def create_no_exist_directory(model_name, ref, tmp_path):
```

MarleneKress79789 (commented):
Looks overall good to me. I did not check the model directory structure, but i don't think its that important because the upload function should work independent of that anyway. and the structure you test seems to have all important cases.

exasol_transformers_extension/udfs/models/model_downloader_udf.py
```diff
@@ -22,11 +24,11 @@ def __init__(self,
     def run(self, ctx) -> None:
         while True:
```
Nicoretti, line 25: a for loop may would be more obvious and less error prone

exasol_transformers_extension/utils/bucketfs_operations.py
```diff
@@ -36,18 +40,24 @@ def create_bucketfs_location(
...
-            file, str(bucketfs_path))
+def upload_file(bucketfs_location: AbstractBucketFSLocation, fileobj: BinaryIO, model_path: Path) -> Path:
+    fileobj.seek(0)
+    model_tar_file = model_path.with_suffix(".tar.gz")
+    bucketfs_location.upload_fileobj_to_bucketfs(fileobj, str(model_tar_file))
+    return model_tar_file
+
+
+def create_tar_of_directory(path: Path, fileobj: BinaryIO):
+    with tarfile.open(name="test.tar.gz", mode="w|gz", fileobj=fileobj) as tar:
```
MarleneKress79789, line 58: maybe rename the test.tar.gz ? this is not a test after all

exasol_transformers_extension/utils/bucketfs_operations.py
```diff
@@ -36,18 +40,24 @@ def create_bucketfs_location(
...
 
 
 @retry(wait=wait_fixed(2), stop=stop_after_attempt(10))
-def upload_model_file(bucketfs_location, model_path, tmp_file_path, tmpdir_name):
-    with open(tmp_file_path, mode='rb') as file:
-        bucketfs_path = PurePosixPath(
-            model_path, tmp_file_path.relative_to(tmpdir_name))
-        bucketfs_location.upload_fileobj_to_bucketfs(
-            file, str(bucketfs_path))
+def upload_file(bucketfs_location: AbstractBucketFSLocation, fileobj: BinaryIO, model_path: Path) -> Path:
```
MarleneKress79789, line 50: here naming also seems inconsistent. i would suggest either changing function name to upload_model or changing the name of model_path (i.e file_path or something)

tests/integration_tests/with_db/test_upload_model.py
```diff
@@ -48,11 +60,7 @@ def test_model_upload(download_sample_models: Path, bucketfs_location: BucketFSL
...
-
-        downloaded_files = set(
-            i.name for i in download_path.iterdir() if i.is_file())
-        uploaded_files = set(
-            i for i in bucketfs_location.list_files_in_bucketfs(upload_path))
-        assert uploaded_files == downloaded_files
+        assert str(upload_path.with_suffix(".tar.gz")) in bucketfs_location.list_files_in_bucketfs(".")
     finally:
-        postprocessing.cleanup_buckets(bucketfs_location, upload_path)
+        pass
```
MarleneKress79789, line 65: is the pass really needed here?

.github/actions/run_unit_tests/action.yml
```diff
@@ -7,10 +7,6 @@ runs:
       run: poetry build
       shell: bash
 
-    - name: Install project whl
-      run: pip install dist/*.whl
```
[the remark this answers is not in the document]
tkilias: pip install needs now needs an additional package index definition, because we fetch the torch package not from pypi, but from torch's package index. Maybe, we should make the torch installation optional for the this package, such that the deployment scripts can be installed without needing to worry about torch.
Another, issue I see with install the whl that way, is that we don't control where we install it, too.

exasol_transformers_extension/udfs/models/model_downloader_udf.py
tkilias: if you can tell me how? The thing is, I would need an until loop which doesn't exist in python.

exasol_transformers_extension/utils/bucketfs_operations.py
tkilias: good catch, actually don't know why I wrote test.tar.gz, maybe copy and paste from a test where I tried this first

tests/integration_tests/with_db/test_upload_model.py
tkilias: the question is actually, can we enable the postprocessing.cleanup again

exasol_transformers_extension/utils/model_downloader.py
```diff
@@ -28,14 +29,25 @@ def __init__(self,
...
     def download_model(self, model_factory: ModelFactoryProtocol):
-        with self._temporary_directory_factory.create() as tmpdir_name:
-            # download model into tmp folder
-            model_factory.from_pretrained(self._model_name, cache_dir=tmpdir_name, use_auth_token=self._token)
+        # download model into tmp folder
+        model_factory.from_pretrained(self._model_name, cache_dir=self._tmpdir_name, use_auth_token=self._token)
 
-            # upload the downloaded model files into bucketfs
-            self._bucketfs_model_uploader.upload_directory(tmpdir_name)
+    def upload_model(self) -> Path:
```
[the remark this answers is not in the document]
tkilias: probably rename class to TransformerHubBucketFSTransfer

tests/integration_tests/with_db/test_upload_model.py
tkilias:
```suggestion
```

exasol_transformers_extension/udfs/models/model_downloader_udf.py
Nicoretti: Something like this:

```python
class ContextIterator:

    def __init__(self, context):
        self._ctx = context

    def __iter__():
        self._ctx.next(reset=True) 

    def __next__():
        self._ctx.next(first=True)
        return self._ctx

    def __getitem__(self, key):
        return self._ctx[key]

    def __getattr__(self, key):
        return self._ctx[key]


def run(self, ctx) -> None:
    for c in ContextIterator(ctx):
        paths = self._download_model(ctx)
        c.emit(*paths)
```

An iterator like this also could be made available in general.

.github/workflows/build.yaml
```diff
@@ -34,3 +34,20 @@ jobs:
 
       - name: Run unit tests
         uses: ./.github/actions/run_unit_tests
```
MarleneKress79789, line 36: maybe add empty line for readability

doc/user_guide/user_guide.md
```diff
@@ -4,6 +4,7 @@
 The Transformers Extension provides a Python library with UDFs that allow the 
 use of pre-trained NLP models provided by the [Transformers API](https://huggingface.co/docs/transformers/index).
 
+The extension provides two types of UDFs:
```
MarleneKress79789, line 7:
```suggestion
```
duplicate line

tests/unit_tests/utils/test_huggingface_hub_bucketfs__model_transfer.py
```diff
@@ -0,0 +1,71 @@
...
+
+from exasol_transformers_extension.utils.bucketfs_model_uploader import BucketFSModelUploader, \
+    BucketFSModelUploaderFactory
+from exasol_transformers_extension.utils.huggingface_hub_bucketfs_model_transfer import ModelFactoryProtocol, \
+    HuggingFaceHubBucketFSModelTransfer
+from exasol_transformers_extension.utils.temporary_directory_factory import TemporaryDirectoryFactory
+from tests.utils.mock_cast import mock_cast
+
+
+class TestSetup:
```
MarleneKress79789, line 15: could this be a fixture instead?

.github/actions/test_pip_install/action.yml
```diff
@@ -0,0 +1,12 @@
...
+runs:
+  using: "composite"
+  steps:
+    - name: Build Poetry
+      run: poetry build
+      shell: bash
+
+    - name: Test pip install
+      run: pip install dist/*.whl --extra-index-url https://download.pytorch.org/whl/cpu
+      shell: bash
```
MarleneKress79789, line 12: git complains about no newline

.github/workflows/build.yaml
tkilias:
```suggestion
        uses: ./.github/actions/run_unit_tests
        
```

.github/actions/test_pip_install/action.yml
tkilias:
```suggestion
      shell: bash
      
```

tests/unit_tests/utils/test_huggingface_hub_bucketfs__model_transfer.py
tkilias: I use this class to group all the necessary mocks. A fixture doesn't allow me this, except if I return, in fact, a class, but then I have a class and a fixture. If we did that, we save one line in the test, but get one more parameter. So, we don't win anything, using a fixture here, because we need a new instance of the class for each test, anyway.

MarleneKress79789 (approved):

PR #111, opened 2023-08-01
#107: Use SLCT api for building the language container

tkilias:


exasol_transformers_extension/deployment/language_container.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+def prepare_flavor(flavor_path: Path):
+    flavor_base_path = flavor_path / "flavor_base"
+    add_wheel_to_flavor(flavor_base_path)
+
+
+def add_wheel_to_flavor(flavor_base_path):
+    dist_path = find_file_or_folder_backwards("dist")
+    subprocess.call(["poetry", "build"])
```
kaklakariada, line 78: This assumes a specific working dir. Maybe explicitly specify the correct working dir

exasol_transformers_extension/deployment/language_container.py
```diff
@@ -0,0 +1,83 @@
...
+    wheels = list(dist_path.glob("*.whl"))
+    wheel = wheels[0]
```
kaklakariada, line 80: check if the correct number of files is found

tests/fixtures/language_container_fixture.py
```diff
@@ -0,0 +1,111 @@
...
+            container_file_path.name)
+    with set_language_alias(flavor_path, itde, container_file_path) as language_alias:
+        wait_for_language_container_ready(itde, language_alias)
+    return container_file_path
+
+
+@pytest.fixture(scope="session")
+def language_alias(itde: TestConfig, flavor_path: Path, upload_slc: Path):
+    with set_language_alias(flavor_path, itde, upload_slc) as language_alias:
+        yield language_alias
```
kaklakariada, line 53: Return type of function is missing? Same for next function

tests/fixtures/language_container_fixture.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+@contextlib.contextmanager
+def set_language_alias(flavor_path: Path, itde: TestConfig, container_file_path: Path):
+    release_name = container_file_path.with_suffix('').with_suffix('').name
```
kaklakariada, line 58:
```suggestion
    release_name = container_file_path.with_suffix('').name
```
❓

tests/fixtures/language_container_fixture.py
```diff
@@ -0,0 +1,111 @@
...
+        itde.ctrl_connection.execute(language_definition.generate_alter_session())
+        itde.ctrl_connection.execute(language_definition.generate_alter_system())
+        yield language_alias
+
+
+def cleanup_images():
+    if "GITHUB_ACTIONS" in os.environ:
+        # Remove image and build output to reduce the disk usage in CI.
+        # We currently, use Github Actions as the CI and its disk is limited to 14 GB.
+        # TODO: This code can be removed if we moved to a CI with larger disks.
```
kaklakariada, line 78:
```suggestion
        # We currently use Github Actions as the CI and its disk is limited to 14 GB.
        # TODO: This code can be removed when we move to a CI with larger disks.
```

tests/fixtures/language_container_fixture.py
```diff
@@ -0,0 +1,111 @@
...
+
+def is_language_container_ready(itde: TestConfig, language_alias: str, schema: str, udf_name: str) -> bool:
+    try:
+        itde.ctrl_connection.execute(textwrap.dedent(f"""
+            CREATE OR REPLACE {language_alias} SCALAR SCRIPT {schema}.{udf_name}(i integer) 
+            RETURNS INTEGER AS
+                def run(ctx):
+                    return 1
+            / 
+            """))
```
kaklakariada, line 106: This creates the UDF every second. Wouldn't it be better to create the UDF once and then just call it every second?

tests/integration_tests/with_db/deployment/test_language_container_deployer.py
```diff
@@ -2,28 +2,59 @@
...
-        language_alias: str,
-        schema: str,
+def test_language_container_deployer(
+        request,
+        export_slc: ExportInfo,
         pyexasol_connection: ExaConnection,
         bucketfs_config: config.BucketFs,
-        container_path,
-        language_settings):
+):
```
kaklakariada, line 20: Type missing for `request` parameter

tests/integration_tests/with_db/deployment/test_language_container_deployer.py
```diff
@@ -2,28 +2,59 @@
...
+def create_and_run_test_udf(pyexasol_connection: ExaConnection, language_alias: str):
+    pyexasol_connection.execute(textwrap.dedent(f"""
+        CREATE OR REPLACE {language_alias} SCALAR SCRIPT "TEST_UDF"()
+        RETURNS BOOLEAN AS
+
+        def run(ctx):
+            return True
+
+        /
+        """))
```
kaklakariada, line 49:
```suggestion
        RETURNS BOOLEAN AS
        def run(ctx):
            return True
        /
        """))
```

tests/integration_tests/with_db/deployment/test_language_container_deployer.py
```diff
@@ -2,28 +2,59 @@
...
+    result = pyexasol_connection.execute('SELECT "TEST_UDF"()').fetchall()
+    return result
```
kaklakariada, line 51: This should always return the same result. You could just
* check the value with `assert result[0][0]`
* don't return the value
* rename the function to `assert_udf_running`

tests/integration_tests/with_db/deployment/test_language_container_deployer_cli.py
```diff
@@ -1,36 +1,45 @@
...
+        CREATE OR REPLACE {language_alias} SCALAR SCRIPT "TEST_UDF"()
+        RETURNS BOOLEAN AS
+    
+        def run(ctx):
+            return True
+    
+        /
+        """))
+        result = pyexasol_connection.execute('SELECT "TEST_UDF"()').fetchall()
+    return result
```
kaklakariada, line 33:
```suggestion
        result = pyexasol_connection.execute('SELECT "TEST_UDF"()').fetchall()
        return result
```
Same here: the result should always be the same. Why return it?

tests/integration_tests/with_db/deployment/test_language_container_deployer_cli.py
```diff
@@ -41,63 +50,57 @@ def _call_deploy_language_container_deployer_cli(
...
-    def run(ctx):
-        return True
 
-    /
-    """))
-    result = db_conn_test.execute('SELECT "TEST_UDF"()').fetchall()
-    return result
+def create_schema(pyexasol_connection: ExaConnection, schema: str):
+    pyexasol_connection.execute(f"DROP SCHEMA IF EXISTS {schema} CASCADE;")
+    pyexasol_connection.execute(f"CREATE SCHEMA IF NOT EXISTS {schema};")
```
kaklakariada, line 73:
```suggestion
    pyexasol_connection.execute(f"CREATE SCHEMA {schema};")
```
Schema should not exist here

tests/integration_tests/with_db/udfs/test_model_downloader_udf_script.py
```diff
@@ -28,13 +25,13 @@ def test_model_downloader_udf_script(
...
-                f"t.token_conn_name"
-                f") FROM (VALUES {str(tuple(input_data))} AS "
-                f"t(model_name, sub_dir, bucketfs_conn_name, token_conn_name));"
+            f"SELECT TE_MODEL_DOWNLOADER_UDF("
+            f"t.model_name, "
+            f"t.sub_dir, "
+            f"t.bucketfs_conn_name, "
+            f"t.token_conn_name"
+            f") FROM (VALUES {str(tuple(input_data))} AS "
+            f"t(model_name, sub_dir, bucketfs_conn_name, token_conn_name));"
```
kaklakariada, line 34: This is hard to read. Could you use a multi-line string here?

tests/utils/revert_language_settings.py
```diff
@@ -1,27 +1,15 @@
...
 
-    return wrapper
+@contextlib.contextmanager
+def revert_language_settings(connection: ExaConnection):
+    language_settings = DBQueries.get_language_settings(connection)
+    try:
+        yield
+    finally:
+        connection.execute(f"ALTER SYSTEM SET SCRIPT_LANGUAGES='{language_settings[0][0]}';")
+        connection.execute(f"ALTER SESSION SET SCRIPT_LANGUAGES='{language_settings[0][1]}';")
```
kaklakariada, line 15: This would be easier to understand if `get_language_settings` would return a class or map with readable property names. It's hard to guess the meaning of `language_settings[0][0]` ;)

kaklakariada (changes requested):

tests/fixtures/language_container_fixture.py
tkilias: no the file is tar.gz file and has with that two suffixes which we need to remove.

tests/integration_tests/with_db/deployment/test_language_container_deployer_cli.py
tkilias: it might be because we don't remove it at the moment

tests/fixtures/language_container_fixture.py
tkilias: With creating it outside of the loop, there is the risk, that the container isn't yet present in the mounted bucketfs filesystem. Which trigger an error during the creation of the UDF.

tests/integration_tests/with_db/deployment/test_language_container_deployer_cli.py
kaklakariada: The schema is dropped literally in the line above ;)

exasol_transformers_extension/deployment/language_container.py
```diff
@@ -19,7 +19,7 @@ def find_file_or_folder_backwards(name: str) -> Path:
     if result_path is not None and result_path.exists():
         return result_path
     else:
-        raise RuntimeError(f"Could not find {name}")
+        raise RuntimeError(f"Could not find {name} when searching backwars from {Path(__file__).parent}")
```
kaklakariada, line 22:
```suggestion
        raise RuntimeError(f"Could not find {name} when searching backwards from {Path(__file__).parent}")
```

exasol_transformers_extension/deployment/language_container.py
```diff
@@ -74,10 +74,14 @@ def prepare_flavor(flavor_path: Path):
...
 def add_wheel_to_flavor(flavor_base_path):
-    dist_path = find_file_or_folder_backwards("dist")
-    subprocess.call(["poetry", "build"])
+    project_directory = find_file_or_folder_backwards("pyproject.toml").parent
+    subprocess.call(["poetry", "build"], cwd=project_directory)
+    dist_path = project_directory / "dist"
     wheels = list(dist_path.glob("*.whl"))
+    if len(wheels) != 1:
+        raise RuntimeError(f"Did not find exactly one wheel file in dist directory {dist_path}. "
+                           f"Found the following wheels {wheels}")
```
kaklakariada, line 83:
```suggestion
                           f"Found the following wheels: {wheels}")
```

tests/utils/revert_language_settings.py
```diff
@@ -11,5 +11,5 @@ def revert_language_settings(connection: ExaConnection):
     try:
         yield
     finally:
-        connection.execute(f"ALTER SYSTEM SET SCRIPT_LANGUAGES='{language_settings[0][0]}';")
-        connection.execute(f"ALTER SESSION SET SCRIPT_LANGUAGES='{language_settings[0][1]}';")
+        connection.execute(f"ALTER SYSTEM SET SCRIPT_LANGUAGES='{language_settings.system_value}';")
+        connection.execute(f"ALTER SESSION SET SCRIPT_LANGUAGES='{language_settings.session_value}';")
```
kaklakariada, line 15: ❤️

kaklakariada (changes requested):

exasol_transformers_extension/utils/bucketfs_operations.py
```diff
@@ -36,11 +38,16 @@ def upload_model_files_to_bucketfs(
...
-                model_path, tmp_file_path.relative_to(tmpdir_name))
-            bucketfs_location.upload_fileobj_to_bucketfs(
-                file, str(bucketfs_path))
+        upload_model_file(bucketfs_location, model_path, tmp_file_path, tmpdir_name)
+
+
+@retry(wait=wait_fixed(2), stop=stop_after_attempt(10))
+def upload_model_file(bucketfs_location, model_path, tmp_file_path, tmpdir_name):
+    with open(tmp_file_path, mode='rb') as file:
+        bucketfs_path = PurePosixPath(
```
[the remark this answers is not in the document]
tkilias: bucketfs_python works with PosixPath, because normal Path allow Windows paths which are not compatible with buckefs

PR #122, opened 2023-08-31
#121: Use matchers for asserts in without db integration tests

### LINKED ISSUE #121: Use matchers in without db integration tests
- currently, the without db integration tests compute many conditions and check in the assert only if these were true, this makes debugging almost impossible, because the pytest output doesn't show what went into the condition
- we can use matcher objects to solve the issue and also remove code duplication between the tests

tkilias:
Fixes #121 

### All Submissions:

*

### DIFF
--- a/tests/integration_tests/without_db/udfs/test_question_answering_udf.py
+++ b/tests/integration_tests/without_db/udfs/test_question_answering_udf.py
@@ -2,6 +2,9 @@ import torch
 import pytest
 import pandas as pd
 from typing import Dict
+
+from tests.integration_tests.without_db.udfs.matcher import Result, ShapeMatcher, NewColumnsEmptyMatcher, \
+    ErrorMessageMatcher, ScoreMatcher, RankDTypeMatcher, NoErrorMessageMatcher, RankMonotonicMatcher, ColumnsMatcher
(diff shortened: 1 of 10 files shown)


tests/integration_tests/without_db/udfs/test_question_answering_udf.py
```diff
@@ -94,27 +96,15 @@ def test_question_answering_udf(
...
-        is_rank_correct
-    ))
+    result = Result(result_df)
+    assert (
+            result == ScoreMatcher()
+            and result == RankDTypeMatcher()
+            and result == RankMonotonicMatcher(n_rows=n_rows, results_per_row=top_k)
+            and result == ShapeMatcher(columns=columns, new_columns=new_columns, n_rows=n_rows, results_per_row=top_k)
+            and result == ColumnsMatcher(columns=columns[1:], new_columns=new_columns)
+            and result == NoErrorMessageMatcher()
```
kaklakariada, line 106: Is this the "pythonic" way of using matchers? Looks a bit strange for Java devs ;)

kaklakariada (changes requested):

kaklakariada (approved):

PR #148, opened 2023-10-27
#124: Better error messages in case of missing parameters

### LINKED ISSUE #124: Add better input validation for better error messages
# Background

- When model_name, bucketfs_conn, or sub_dir are NULL the function extract_unique_model_dataframes_from_batch returns an empty dataframe which will lead to the following error message `RuntimeError: E-UDF-CL-SL-PYTHON-1087: emit DataFrame is empty`
- model_name, bucketfs_conn, or sub_dir can't be really empty, for that reason failing is actually correct, but we should give a better error message what is wrong

MarleneKress79789:
### All Submissions:

Fixes #124

### DIFF
--- a/tests/unit_tests/udfs/base_model_dummy_implementation.py
+++ b/tests/unit_tests/udfs/base_model_dummy_implementation.py
@@ -20,7 +20,6 @@ class DummyImplementationUDF(BaseModelUDF):
 
     def extract_unique_param_based_dataframes(
             self, model_df: pd.DataFrame) -> Iterator[pd.DataFrame]:
-
         yield model_df
 
     def execute_prediction(self, model_df: pd.DataFrame) -> \
@@ -32,7 +31,6 @@ class DummyImplementationUDF(BaseModelUDF):
             self, model_df: pd.DataFrame, pred_df_list: List[pd.DataFrame]) \
             -> pd.DataFrame:
 
-        # Concat predictions and model_df
         pred_df = pd.concat(pred_df_list, axis=0).reset_index(drop=True)
         model_df = pd.concat([model_df, pred_df], axis=1)
         return model_df
--- a/tests/unit_tests/udfs/base_model_dummy_implementation.py
+++ b/tests/unit_tests/udfs/base_model_dummy_implementation.py
@@ -1,0 +1,54 @@
+
+import pandas as pd
+import transformers
+from typing import List, Iterator, Any, Dict, Union
+from exasol_transformers_extension.udfs.models.base_model_udf import \
+    BaseModelUDF
+
+
+class DummyImplementationUDF(BaseModelUDF):
+    def __init__(self,
+                 exa,
+                 batch_size=100,
+                 pipeline=transformers.pipeline,
+                 base_model=transformers.AutoModel,
+                 tokenizer=transformers.AutoTokenizer):
+        super().__init__(exa, batch_size, pipeline, base_model,
+                         tokenizer, 'dummy_task')
+        self._desired_fields_in_prediction = ["answer", "score"]
+        self.new_columns = ["answer", "score", "error_message"]
+
+    def extract_unique_param_based_dataframes(
+            self, model_df: pd.DataFrame) -> Iterator[pd.DataFrame]:
+
+        yield model_df
+
+    def execute_prediction(self, model_df: pd.DataFrame) -> \
+            List[Union[Dict[str, Any], List[Dict[str, Any]]]]:
+        dummy_result = [{"answer": True, "score": "1"}]
+        return dummy_result
+
+    def append_predictions_to_input_dataframe(
+            self, model_df: pd.DataFrame, pred_df_list: List[pd.DataFrame]) \
+            -> pd.DataFrame:
+
+        # Concat predictions and model_df
+        pred_df = pd.concat(pred_df_list, axis=0).reset_index(drop=True)
+        model_df = pd.concat([model_df, pred_df], axis=1)
+        return model_df
+
+    def create_dataframes_from_predictions(
+            self, predictions: List[Union[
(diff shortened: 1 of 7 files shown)


tests/unit_tests/udfs/base_model_dummy_implementation.py
```diff
@@ -0,0 +1,54 @@
...
+                 base_model=transformers.AutoModel,
+                 tokenizer=transformers.AutoTokenizer):
+        super().__init__(exa, batch_size, pipeline, base_model,
+                         tokenizer, 'dummy_task')
+        self._desired_fields_in_prediction = ["answer", "score"]
+        self.new_columns = ["answer", "score", "error_message"]
+
+    def extract_unique_param_based_dataframes(
+            self, model_df: pd.DataFrame) -> Iterator[pd.DataFrame]:
+
```
tkilias, line 23:
```suggestion

```

tests/unit_tests/udfs/base_model_dummy_implementation.py
```diff
@@ -0,0 +1,54 @@
...
+    def execute_prediction(self, model_df: pd.DataFrame) -> \
+            List[Union[Dict[str, Any], List[Dict[str, Any]]]]:
+        dummy_result = [{"answer": True, "score": "1"}]
+        return dummy_result
+
+    def append_predictions_to_input_dataframe(
+            self, model_df: pd.DataFrame, pred_df_list: List[pd.DataFrame]) \
+            -> pd.DataFrame:
+
+        # Concat predictions and model_df
```
tkilias, line 35:
```suggestion
```

I think the comment isn't so useful here

tests/unit_tests/udfs/test_base_udf.py
```diff
@@ -0,0 +1,141 @@
...
+
+    udf.run(mock_ctx)
+    res = mock_ctx.output
+    print(res)
+    error_field = res[0][-1]
+    expected_error_start = f"For each model model_name, bucketfs_conn and sub_dir need to be provided. " \
+                           f"Found model_name = {model_name},"
+    expected_error_end = f" sub_dir = {sub_dir}"
+    assert error_field is not None and len(res[0]) == len(mock_meta.output_columns)
+    assert expected_error_start in error_field and expected_error_end in error_field
```
tkilias, line 141: you could use regex_matcher , not sure if we have it already in this project, otherwise take a look in other projects

tkilias (changes requested):

tests/unit_tests/udfs/test_base_udf.py
```diff
@@ -0,0 +1,141 @@
...
+        None)
+    mock_ctx = create_mock_udf_context(input_data, mock_meta)
+
+    udf = DummyImplementationUDF(exa=mock_exa,
+                                 base_model=mock_base_model_factory.side_effect,
+                                 tokenizer=mock_tokenizer_factory.side_effect)
+
+    udf.run(mock_ctx)
+    res = mock_ctx.output
+    print(res)
```
MarleneKress79789, line 135:
```suggestion
```

tests/unit_tests/udfs/test_base_udf.py
```diff
@@ -133,8 +134,8 @@ def test_model_downloader_missing_parameters(description, bucketfs_conn_name, bu
     udf.run(mock_ctx)
     res = mock_ctx.output
     error_field = res[0][-1]
-    expected_error_start = f"For each model model_name, bucketfs_conn and sub_dir need to be provided. " \
-                           f"Found model_name = {model_name},"
-    expected_error_end = f" sub_dir = {sub_dir}"
+    pattern = f"For each model model_name, bucketfs_conn and sub_dir need to be provided. Found model_name = " \
+              f"{model_name}, bucketfs_conn = .*, sub_dir = {sub_dir}."
+
+    assert re.search(pattern, error_field)
```
tkilias, line 140: I meant something, like that. It gives you a nicer diff

https://github.com/exasol/integration-test-docker-environment/blob/d3450d947ea88e0d4900beac2a2a0cfd65c3806a/test/matchers.py#L4

tests/unit_tests/udfs/test_base_udf.py
```diff
@@ -0,0 +1,141 @@
...
+            Column("answer", bool, "BOOLEAN"),
+            Column("score", str, "DOUBLE"),
+            Column("error_message", str, "VARCHAR(2000000)")
+        ]
+    )
+    return meta
+
+
+@pytest.mark.parametrize("description, bucketfs_conn_name, bucketfs_conn ,"
+                         "sub_dir, model_name", [
```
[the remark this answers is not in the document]
tkilias: When I look at the tests, you probably, can also simplify them, because they share a lot of code, almost everything except input_data construction and asserts, so maybe extract that.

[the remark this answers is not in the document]
tkilias: or alternatively, I sometimes have only one test and a parameter which specifies, if the test is expected to produce an error, and have two different asserts depending on that parameter

tkilias (changes requested):

tests/unit_tests/udfs/test_base_udf.py
```diff
@@ -14,6 +14,30 @@
...
+    def __repr__(self):
+        return self._regex.pattern
+
+
+class regex_matcher2: #this one looks nicer when used and gives better error messages
+    """Assert that a given string meets some expectations."""
+    def __init__(self, some_sting):
+        self._string = some_sting
+
+    def __contains__(self, pattern):
```
tkilias, line 34: The __eq__ in the linked example has a reason.  This gives you a good diff. You used it then

`assert actual_value == regex_matcher("abc.*")`

tkilias (changes requested):

tkilias (approved):

Issue #133, opened 2023-10-10
Improve documentation with missing info and better readability


MarleneKress79789:
```[tasklist]
### Tasks
- [x] overhaul documentation for better readability
- [x] do we want to use pyexasol for the connection in the user guide? looks cleaner imo
- [x] also info about adding SSL certificate?
- [x] mention each model is 2 part model and transformer (in dev guide at least)
- [ ] in dev docu: add info about change of model if parameters changed(happens in base_model_udf), and how
- [ ] in user docu: mention udf can change udf/model if the input set changes the model parameters
- [ ] add example usages in user docu (will be done in #164)
- [x] mention trigger slc dowload test after release to make sure it works with new version. test_language_container_deployer_cli_by_downloading_container (maybe trigger automatically in nox session which does a release)
- [x] add information about the certificates as explained [here](https://github.com/exasol/transformers-extension/pull/150/files/99549de939e855f1698a05a4145d3373f51ffab9#r1392322048 )
- [x] add explanation about model download and uplaod, details [here](https://github.com/exasol/transformers-extension/pull/150#discussion_r1392443311)
- [x] mention after release check .dist and .dist in language container for multiple wheels
- [x] mention windows not supported
- [x] change description of project to mention Huggingface
- [x] see comments in https://github.com/exasol/transformers-extension/pull/150
- [x] link to udf docu
- [x] link to bucketfs docu
- [ ] add screenshot from exa operations where to find parameters like bucketfs name (see screenshot) #164
- [x] mention pypi release and pip install
- [x] get correct link  to pypi and install comand
```

ahsimb:
Please add some explanation for the Model Uploader Script. In particular we need to mention that the `--local-model-path` needs to point to a cache directory. For example, one may download a model using this code:
```
from transformers import AutoTokenizer, AutoModelForSequenceClassification
tokenizer = AutoTokenizer.from_pretrained(<MODEL-NAME>, cache_dir="whatever")
model = AutoModelForSequenceClassification.from_pretrained(<MODEL-NAME>,, cache_dir="whatever")
```
Then it can be uploaded to the BucketFS with this command:
```
python -m exasol_transformers_extension.upload_model \
   ...
    --local-model-path whatever     
```

ahsimb:
[...]

PR #150, opened 2023-11-13
#132: Fixed outdated documentation

### LINKED ISSUE #132: Fix wrong information in Documentation
```[tasklist]
### Tasks
- [x] in dev docu: build and then download build archive is very strange. remove mention of download
- [x] need to updated dev docu: run tests should now be done using nox, also needs to mention starting db
- [x] user guide: "python -m exasol_transformers_extension.deploy language-container --version 0.5.0 ..." tries to download slc from github but does not seem to exist (might be a wrong name issue)
- [x] deployment.deployment_utils.py : should be: GH_RELEASE_URL = "https://github.com/exasol/transformers-extension/releases/tag" (No this is correct as is)
```

MarleneKress79789:
### All Submissions:

Fixes #132

### DIFF
--- a/doc/developer_guide/developer_guide.md
+++ b/doc/developer_guide/developer_guide.md
@@ -5,9 +5,10 @@ In this developer guide we explain how to build this project and how you can add
 new transformer tasks and tests.
 
 
-## Building the Project
+## Installation
+There are two ways to install the Transformers Extension Package:
 
-### 1. Build the Python Package
+### 1. Build and install the Python Package
 This project needs Python 3.8 or above installed on the development machine. 
 In addition, in order to build Python packages you need to have the [Poetry](https://python-poetry.org/)
 (>= 1.1.11) package manager. Then you can install and build the `transformers-extension` as follows:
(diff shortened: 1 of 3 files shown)


doc/developer_guide/developer_guide.md
ahsimb, line 1: We need to elaborate on how to upload a model by downloading it from the Huggingface Hub to a local drive.

Firstly, a sample code like the one below would be useful. This code will create a model cache. This cache is actually what we need to upload to the BucketFS.
```
from transformers import AutoTokenizer, AutoModel

AutoTokenizer.from_pretrained(model_name, cache_dir=model_dir, token=user_token)
AutoModel.from_pretrained(model_name, cache_dir=model_dir, token=user_token)
```

For example, if `model_name == 'me/my-awesome-model'`, and `model_dir == 'my_model_dir'` then the above code will create some files in the directory `'.../my_model_dir/models--me--my-awesome-model'`.

The `'.../my_model_dir'` is what we need to provide to the `exasol_transformers_extension.upload_model` in the `local-model-path` parameter. However, and this is important, it will grab EVERYTHING that it finds in this directory. If you downloaded several models there it will wrap them all in a single tar.gz file and start uploading. Therefore it is important to cache every model in its individual sub-directory. For example, if the root cache directory is `'my-models-cache'` then I shall set the `model_dir` to something like `'.../my-models-cache/my-awesome-model-cache'`.

MarleneKress79789: Yes, you already mentioned that in #133. i don´t think it is relevant to this ticket, so i would suggest working on adding missing information to the documentation in #133 since that is the focus of that ticket.

doc/user_guide/user_guide.md
```diff
@@ -107,8 +109,7 @@ deployment script below with the desired version. (see GitHub Releases
       --language-alias <LANGUAGE_ALIAS> \ 
       --version <RELEASE_VERSION> \
       --ssl-cert-path <ssl-cert-path> \
-      --use-ssl-cert-validation \
-      --no-use-ssl-cert-valiation
+      --use-ssl-cert-validation
   ```
 The `--ssl-cert-path` is optional if your certificate is not in the OS truststore.
```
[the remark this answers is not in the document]
MarleneKress79789: i will add it to #133.

ahsimb (approved):

Issue #145, opened 2023-10-20
Add load class which loads model from local path


MarleneKress79789:
# Background:

- Currently, we use from_pretrained with the cache_dir option to load the model and upload the cache directory to the BucketFS
- The current approach in general works, but seems to check each time online if the model exists, or we have access and if we have the newest version
- This is not possible in some deployments 
- To solve this issue, we could follow the following [example](https://huggingface.co/docs/transformers/v4.32.1/en/installation#fetch-models-and-tokenizers-to-use-offline) using save_pretrained which stores the model into a directory which we then could upload
- We can then load the model with from_pretrained from a local path
- After we extracted the load class in #144 , we can then create a new load function which loads the model from a local path during inference

```[tasklist]
### Acceptance Criteria
- [x] Create a load class, which uses a local path to load the model during inference
- [x] add tests for new load function
- [x] add documentaion for the new load function like doc strings, full user docu will be added in #146
```

ckunki:
Did some review, see comments in Confluence. 
Effort: 0.1

PR #161, opened 2023-12-01
#144: Extracted load_models into seperate class

### LINKED ISSUE #144: Extract base_model_udf.load_models into separate class
# Background:

- Currently, we use from_pretrained with the cache_dir option to load the model and upload the cache directory to the BucketFS
- The current approach in general works, but seems to check each time online if the model exists,  we have access and  we have the newest version
- This is not possible in some deployments 
- To solve this issue, we could follow the following [example](https://huggingface.co/docs/transformers/v4.32.1/en/installation#fetch-models-and-tokenizers-to-use-offline) using save_pretrained which stores the model into a directory which we then could upload
- We can then load the model with from_pretrained from a local path
- It would be easier to change the model loading once it is extracted into its own class

# Acceptance Criteria

- [ ]  Extract [base_model_udf.load_models](https://github.com/exasol/transformers-extension/blob/1321af15f1cf0e9ac68ca12c02a3d6e4318a2cb1/exasol_transformers_extension/udfs/models/base_model_udf.py#L187C13-L187C13) into seperate class and inject it into base_model_udf

MarleneKress79789:
### All Submissions:

Fixes #144

### DIFF
--- a/exasol_transformers_extension/udfs/models/base_model_udf.py
+++ b/exasol_transformers_extension/udfs/models/base_model_udf.py
@@ -53,12 +53,11 @@ class BaseModelUDF(ABC):
             predictions_df = self.get_predictions_from_batch(batch_df)
             ctx.emit(predictions_df)
 
-        self.clear_device_memory()
+        self.model_loader.clear_device_memory()
 
     def create_model_loader(self):
         """
-        creates the model_loader. In separate function, so it can be replaced for tests since the pipeline
-        creation does not work with dummy data
+        Creates the model_loader.
         """
         self.model_loader = LoadModel(self.pipeline,
                                       self.base_model,
@@ -183,7 +182,7 @@ class BaseModelUDF(ABC):
         current_model_key = (bucketfs_conn, sub_dir, model_name, token_conn)
         if self.model_loader.last_loaded_model_key != current_model_key:
             self.set_cache_dir(model_name, bucketfs_conn, sub_dir)
-            self.clear_device_memory()
+            self.model_loader.clear_device_memory()
             if token_conn:
                 token_conn_obj = self.exa.get_connection(token_conn)
             else:
@@ -211,13 +210,6 @@ class BaseModelUDF(ABC):
         self.cache_dir = bucketfs_operations.get_local_bucketfs_path(
             bucketfs_location=bucketfs_location, model_path=str(model_path))
 
-    def clear_device_memory(self):
-        """
-        Delete models and free device memory
-        """
-        self.model_loader.last_loaded_model = None
-        self.model_loader.last_loaded_tokenizer = None
-        torch.cuda.empty_cache()
 
     def get_prediction(self, model_df: pd.DataFrame) -> pd.DataFrame:
         """
--- a/exasol_transformers_extension/udfs/models/base_model_udf.py
+++ b/exasol_transformers_extension/udfs/models/base_model_udf.py
@@ -37,6 +37,7 @@ class BaseModelUDF(ABC):
         self.device = None
         self.cache_dir = None
         self.model_loader = None
+        self.last_created_pipeline = None
         self.new_columns = []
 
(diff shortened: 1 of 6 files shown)


exasol_transformers_extension/udfs/models/base_model_udf.py
```diff
@@ -199,36 +215,10 @@ def clear_device_memory(self):
         """
         Delete models and free device memory
         """
-        self.last_loaded_model = None
-        self.last_loaded_tokenizer = None
+        self.model_loader.last_loaded_model = None
+        self.model_loader.last_loaded_tokenizer = None
```
ahsimb, line 219: I would add a `clear` method to the ModelLoader setting these two variables to None.
If `last_laded_model` and `last_loaded_tokenizer` still need to be visible I would make them properties.

BTW, if the `last_created_pipeline` keeps references to these objects they won't be garbage-collected. So you probably need to it None too.

tests/unit_tests/udfs/base_model_dummy_implementation.py
```diff
@@ -6,6 +6,38 @@
...
+                    cache_dir,
+                    token_conn_obj) -> None:
+        token = False
+        self.last_loaded_model = self.base_model.from_pretrained(
+            model_name, cache_dir=cache_dir, use_auth_token=token)
+        self.last_loaded_tokenizer = self.tokenizer.from_pretrained(
+            model_name, cache_dir=cache_dir, use_auth_token=token)
+        return None
+
+
```
ahsimb, line 40: Why can't you use the actual ModelLoader with a pipeline-like function that does nothing and returns None?

exasol_transformers_extension/udfs/models/base_model_udf.py
MarleneKress79789: will move the clear method from base udf also

exasol_transformers_extension/utils/load_model.py
```diff
@@ -0,0 +1,48 @@
...
+        self.task_name = task_name
+        self.device = device
+        self.last_loaded_model = None
+        self.last_loaded_tokenizer = None
+        self.last_loaded_model_key = None
+
+    def load_models(self, model_name: str,
+                    current_model_key,
+                    cache_dir,
+                    token_conn_obj) -> None:
```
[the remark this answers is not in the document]
MarleneKress79789: regarding the documentation strings and stuff: this whole class will be replaced with the one that will be created in #145. so i would rather spend the time to do proper docu for the new class and leave this one as is.

tests/unit_tests/udfs/base_model_dummy_implementation.py
MarleneKress79789: You are completely right, i misunderstood. changed

exasol_transformers_extension/utils/load_model.py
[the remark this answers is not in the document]
ahsimb: I would say the released code should be in a state of completeness. If the class morphs into something else so will the documentation. Plus it doesn't take long to get the docstrings sorted.

ahsimb (approved):

exasol_transformers_extension/utils/load_model.py
```diff
@@ -0,0 +1,48 @@
...
+        Load model and tokenizer model from the cached location in bucketfs.
+        If the desired model is not cached, this method will attempt to
+        download the model to the read-only path /bucket/.. and cause an error.
+        This error will be addressed in ticket
+        https://github.com/exasol/transformers-extension/issues/43.
+
+        :param model_name: The model name to be loaded
+        """
+        token = False
+        if token_conn_obj:
```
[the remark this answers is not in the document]
MarleneKress79789: new bug ticket added here: https://github.com/exasol/transformers-extension/issues/163

PR #265, opened 2024-10-15
#264 Created a single deploy command

### LINKED ISSUE #264: Made a single installation CLI command using the features in PEC
This should be similar to the [installation command](https://github.com/exasol/sagemaker-extension/blob/main/exasol_sagemaker_extension/deploy.py) implemented in the SageMaker extension. In the case of the Transformers Extension it should also include the creation of the BucketFS connection object.

ahsimb:
closes #264 

### All Submissions:


exasol_transformers_extension/deploy.py
```diff
@@ -1,29 +1,86 @@
...
+opts.append(click.Option([get_opt_name(BUCKETFS_CONN_NAME_ARG)], type=str))
+opts.append(click.Option([get_opt_name(TOKEN_CONN_NAME_ARG)], type=str))
+opts.append(click.Option([get_opt_name(TOKEN_ARG)], type=str))
+
+
+def deploy(**kwargs):
+
+    # Make sure there is a valid language_alias
+    if not kwargs.get(StdParams.language_alias.name):
+        kwargs[StdParams.language_alias.name] = 'PYTHON3_TE'
```
ckunki, line 51: This may indicate an optional future improvement, i.e. specifying application-default values for specific CLI commands. Or maybe variable `opts` above allows to modify default values, too?

exasol_transformers_extension/deploy.py
```diff
@@ -1,29 +1,86 @@
...
+    opt_name = arg_name.replace("_", "-")
+    return f'--{opt_name}/--no-{opt_name}'
+
+
+opts = select_std_options([StdTags.DB, StdTags.BFS, StdTags.SLC], formatters=formatters)
+opts.append(click.Option([get_bool_opt_name(DEPLOY_SLC_ARG)], type=bool, default=True))
+opts.append(click.Option([get_bool_opt_name(DEPLOY_SCRIPTS_ARG)], type=bool, default=True))
+opts.append(click.Option([get_opt_name(BUCKETFS_CONN_NAME_ARG)], type=str))
+opts.append(click.Option([get_opt_name(TOKEN_CONN_NAME_ARG)], type=str))
+opts.append(click.Option([get_opt_name(TOKEN_ARG)], type=str))
```
ahsimb, line 44: This option must be made a secret

tests/integration_tests/with_db/deployment/test_deploy_cli.py
```diff
@@ -0,0 +1,64 @@
...
+    deploy_command, DEPLOY_SLC_ARG, BUCKETFS_CONN_NAME_ARG, get_bool_opt_name)
+from exasol_transformers_extension.deployment.language_container import export_slc
+from tests.integration_tests.with_db.test_upload_model import run_model_upload_test
+
+PATH_IN_BUCKET = 'te_end2end'
+BUCKETFS_CONN_NAME = 'TE_E2E_BFS_CONN'
+LANGUAGE_ALIAS = 'TE_E2E_LANG_ALIAS'
+
+
+@pytest.mark.skip('Need to sort out the model upload test first')
```
ckunki, line 19: Will the comment be resolved in a different ticket resulting in the current tests to be no longer skipped?

tests/integration_tests/with_db/deployment/test_deploy_cli.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+def test_scripts_deployer_cli_with_encryption_verify(backend,
+                                                     pyexasol_connection,
+                                                     language_alias,
+                                                     cli_args,
+                                                     deployed_slc):
+    if backend != BACKEND_ONPREM:
+        pytest.skip(("We run this test only with the Docker-DB "
+                     "because SaaS always verifies the SSL certificate"))
```
ckunki, line 50: Isn't there a fixture that will run the current test only only onprem?

tests/integration_tests/with_db/deployment/test_deploy_cli.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+@pytest.mark.skip('Need to sort out the model upload test first')
+def test_deploy_cli(pyexasol_connection,
+                    backend_aware_database_params,
+                    backend_aware_bucketfs_params,
+                    bucketfs_cli_args,
+                    cli_args):
+
+    with temp_schema(pyexasol_connection) as schema_name:
```
ckunki, line 26: Would it make sense to create a fixture `temp_schema`?

ckunki (changes requested):
Added some comments / questions / proposals

exasol_transformers_extension/deploy.py
ahsimb: Good point. I can actually override the definition of a standard option. The deficiency in PEC I can see now is that it simply replaces one definition with another. It could be smarter and merge the two definitions. For example, here I just want to override the default value. I don't want to change the help string etc.

tests/integration_tests/with_db/deployment/test_deploy_cli.py
ahsimb: Spot on!!! I forgot to remove it. Which means I haven't actually run this test. Verey embarrassing.

tests/integration_tests/with_db/deployment/test_deploy_cli.py
ahsimb: This test checks that we will get a certain SSL error if we request the server certificate verification when the server can't provide it. In my opinion, this test is unnecessary as it actually tests the pyexasol and nothing specific to TE. With your blessing, I will remove this test.

tests/integration_tests/with_db/deployment/test_deploy_cli.py
ahsimb: I don't think so. Why would we need a fixture?

exasol_transformers_extension/deploy.py
```diff
@@ -36,20 +36,24 @@ def get_bool_opt_name(arg_name: str) -> str:
...
-opts.append(click.Option([get_bool_opt_name(DEPLOY_SLC_ARG)], type=bool, default=True))
-opts.append(click.Option([get_bool_opt_name(DEPLOY_SCRIPTS_ARG)], type=bool, default=True))
-opts.append(click.Option([get_opt_name(BUCKETFS_CONN_NAME_ARG)], type=str))
-opts.append(click.Option([get_opt_name(TOKEN_CONN_NAME_ARG)], type=str))
-opts.append(click.Option([get_opt_name(TOKEN_ARG)], type=str))
+opt_lang_alias = {'type': str, 'default': 'PYTHON3_TE'}
+opt_token = {'type': str, help: 'Huggingface token for private models'}
+make_option_secret(opt_token, prompt='Huggingface token')
+opts = select_std_options([StdTags.DB, StdTags.BFS, StdTags.SLC],
+                          formatters=formatters, override={StdParams.language_alias: opt_lang_alias})
```
ckunki, line 43: :+1:

tests/integration_tests/with_db/deployment/test_deploy_cli.py
ckunki: We don't need a fixture, but it could be beneficial IMHO
* Using the fixture `temp_schema` can replace using fixture `pyexasol_connection` 
* and also replaces the `with` statement in the test.
* reduces the indent level in the test case

tests/integration_tests/with_db/deployment/test_deploy_cli.py
ckunki: I am happy to remove this test. :slightly_smiling_face:

exasol_transformers_extension/deploy.py
ckunki: :+1:

ckunki (approved):

tests/integration_tests/with_db/deployment/test_deploy_cli.py
ahsimb: We can consider adding a test-level fixture `temp-schema` or a pyexasol connection opened with a temporary schema into the pytest plugins. For now, I think it's alright as it is.

doc/user_guide/user_guide.md
```diff
@@ -290,36 +257,24 @@ severely. Available task_types are the same as the names of our available UDFs,
...
-| [no_]use-ssl-cert-validation |   [x]   | [x]  | Optional boolean, defaults to True              |
+You can invoke the Python script below which downloads the transformer 
+models from The Hugging Face hub to the local filesystem, then to the BucketFS.
+
+```shell
+python -m exasol_transformers_extension.upload_model <options>
+```
+
+For information about the available options common to all Exasol extensions please refer to the
+[documentation](https://github.com/exasol/python-extension-common/blob/0.8.0/doc/user_guide/user-guide.md) in the Exasol Python Extension Common package.
```
ckunki, line 268: I'm not 100% sure about the versioned references to the PEC user guide.

I see the risk that in future we might forget to update these references.
Maybe it would be risky as well but still better to fall back to the latest version `PEC/doc/user_guide/user-guide.md`?

As a minimal mitigation, we could also define a variable (e.g. at the beginning of the file) and use that to create identical links in both places:
```
[pec-user-guide]: https://github.com/exasol/python-extension-common/blob/0.8.0/doc/user_guide/user-guide.md
<!-- Please check for new releases and update the reference to point to the latest version, if appropriate -->

[documentation][pec-user-guide]
```

tests/integration_tests/with_db/deployment/test_deploy_cli.py
ckunki: :+1:

doc/user_guide/user_guide.md
```diff
@@ -56,8 +56,12 @@ models on the Exasol Cluster. More information on The BucketFS can be found
 - The Exasol cluster must already be running with version 7.1 or later.
 - DB connection information and credentials are needed.
 
-### BucketFS Connection 
-An Exasol connection object must be created with Exasol BucketFS connection information and credentials. 
+### BucketFS Connection
+An Exasol connection object must be created with Exasol BucketFS connection information and credentials.
+Normally, the connection object is created as part of the Transformers Extension deployment 
+(see the [Setup section](#deploy-the-extension-to-the-database) below). This section describes how this object
+can be created manually.
```
ckunki, line 63: To be picky: each heading should be followed by an empty line.

Additionally, in the past we found out, that paragraphs split into multiple lines can small updates (e.g. inserting a single word) to potentially cause a diff affecting multiple lines. Hence, we usually prefer to format paragraphs in a single source line only. 

I personally must admit, that multiple lines are much more pleasant to read in the editor :slightly_frowning_face:

ckunki (dismissed):

tests/integration_tests/with_db/deployment/test_scripts_deployer.py
```diff
@@ -14,25 +14,23 @@
...
 def test_scripts_deployer(
         backend,
-        deploy_params: dict[str, Any],
+        database_std_params,
         pyexasol_connection: ExaConnection,
         language_alias,
         deployed_slc):
 
     with temp_schema(pyexasol_connection) as schema_name:
         # We validate the server certificate in SaaS, but not in the Docker DB
```
MarleneKress79789, line 23: this comment seems left over

doc/user_guide/user_guide.md
```diff
@@ -290,36 +257,24 @@ severely. Available task_types are the same as the names of our available UDFs,
...
-| saas-database-name           |         | [x]  | Optional, provide if the database_id is unknown |
-| saas-token                   |         | [x]  | Env. [SAAS_TOKEN]                               |
-| model-name                   |   [x]   | [x]  |                                                 |
-| path-in-bucket               |   [x]   | [x]  | Root location in the bucket for all models      |
-| sub-dir                      |   [x]   | [x]  | Sub-directory where this model should be stored |
-| task_type                    |   [x]   | [x]  | Name of the task you want to use the model for  |
-| token                        |   [x]   | [x]  | Huggingface token (needed for private models)   |
-| [no_]use-ssl-cert-validation |   [x]   | [x]  | Optional boolean, defaults to True              |
+You can invoke the Python script below which downloads the transformer 
+models from The Hugging Face hub to the local filesystem, then to the BucketFS.
```
MarleneKress79789, line 261: I prefer the existing sentence here. the script exists specifically for people who do not want to connect from their DB directly to the huggingface hub, but rather want to only connect to the db from their own machine. i think this should be empathized.

exasol_transformers_extension/deploy.py
```diff
@@ -1,29 +1,90 @@
...
+ver_formatter.set_formatter(CONTAINER_NAME_ARG, TeLanguageContainerDeployer.SLC_NAME)
+formatters = {StdParams.version: ver_formatter}
 
 
-slc_parameter_formatters.set_formatter(CustomizableParameters.container_url,
-                                       TeLanguageContainerDeployer.SLC_URL_FORMATTER)
-slc_parameter_formatters.set_formatter(CustomizableParameters.container_name,
-                                       TeLanguageContainerDeployer.SLC_NAME)
+def get_opt_name(arg_name: str) -> str:
+    # This and the next function should have been implemented in the PEC.
```
MarleneKress79789, line 30: maybe create a ticket to move them then?

exasol_transformers_extension/deploy.py
```diff
@@ -1,29 +1,90 @@
...
+
+
+opt_lang_alias = {'type': str, 'default': 'PYTHON3_TE'}
+opt_token = {'type': str, 'help': 'Huggingface token for private models'}
+make_option_secret(opt_token, prompt='Huggingface token')
+opts = select_std_options([StdTags.DB, StdTags.BFS, StdTags.SLC],
+                          formatters=formatters, override={StdParams.language_alias: opt_lang_alias})
+opts.append(click.Option([get_bool_opt_name(DEPLOY_SLC_ARG)], type=bool, default=True,
+                         help='Deploy SLC'))
+opts.append(click.Option([get_bool_opt_name(DEPLOY_SCRIPTS_ARG)], type=bool, default=True,
```
MarleneKress79789, line 46: i am curious. why do you do the arguments like this instead of a standard click option?

doc/user_guide/user_guide.md
ahsimb: Referencing a particular version of PEC was precisely the intention here. This is the way to resolve conflicts when the newest version of the PEC documentation describes things that are not relevant to the latest version of the extension.
However, you've made a good point that the references are buried somewhere in the doc (and there are two of them making it worse), which makes it very easy to forget to update them. I will try your suggestion.

doc/user_guide/user_guide.md
ahsimb: "each heading should be followed by an empty line." I can't see a difference in the viewer on my screen.

tests/integration_tests/with_db/deployment/test_deploy_cli.py
```diff
@@ -0,0 +1,41 @@
...
+LANGUAGE_ALIAS = 'TE_E2E_LANG_ALIAS'
+
+
+def test_deploy_cli(pyexasol_connection,
+                    backend_aware_database_params,
+                    backend_aware_bucketfs_params,
+                    bucketfs_cli_args,
+                    cli_args):
+    """
+    This is an end-to-end test that should be moved into a separate group of tests.
```
MarleneKress79789, line 24: i dont really like this comment here. can you remove it and create a ticket instead?

doc/user_guide/user_guide.md
ahsimb: The existing sentence doesn't say this either. I will add your explanations of who this script is for.

ahsimb: Actually, we are already saying this in the starting paragraph of the **Store Models in BucketFS** section.

"The Model Downloader UDF is the simpler option, but if you do not want to connect your Exasol Database
directly to the internet, the Model Uploader Script is an option for you."

There is no need to repeat this.

exasol_transformers_extension/deploy.py
ahsimb: The answer depends on what you mean by a standard click option.
We could have added these two to the `StdParams` in PEC, if you are referring to this kind of standard options. At least the first one - `deploy_slc`. If there is an SLC the must be a way to skip its deployment. The other one - `deploy_scripts` - I am not sure. It doesn't fit into any group. We can have an extension without any scripts, in theory.

tests/integration_tests/with_db/deployment/test_deploy_cli.py
ahsimb: OK, will do.

ckunki (approved):

PR #277, opened 2024-11-21
Bug/#275 fix error on no token found

MarleneKress79789:
### All Submissions:

### DIFF
--- a/exasol_transformers_extension/udfs/models/token_classification_udf.py
+++ b/exasol_transformers_extension/udfs/models/token_classification_udf.py
@@ -120,7 +120,8 @@ class TokenClassificationUDF(BaseModelUDF):
 
         # Concat predictions and model_df
         pred_df = pd.concat(pred_df_list, axis=0).reset_index(drop=True)
-        model_df = pd.concat([model_df, pred_df], axis=1)
+        model_df = pd.concat([model_df, pred_df], axis=1, join='inner') # join='inner' -> drop rows where results are empty
+
         if self.work_with_spans:
             model_df = self.create_new_span_columns(model_df)
             model_df[["entity_docid", "entity_char_begin", "entity_char_end"]] =\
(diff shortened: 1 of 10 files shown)


exasol_transformers_extension/udfs/models/token_classification_udf.py
```diff
@@ -141,12 +142,24 @@ def create_dataframes_from_predictions(
...
         results_df_list = []
         for result in predictions:
-            result_df = pd.DataFrame(result)
-            result_df = result_df[self._desired_fields_in_prediction].rename(
-                columns={
-                    "start": "start_pos",
-                    "end": "end_pos",
-                    "entity_group": "entity"})
+            if result:
+                result_df = pd.DataFrame(result)
```
ahsimb, line 146: According to the type hint, the `result` can be either `List[Dict[str, Any]]`, in which case it will work, or `Dict[str, Any]`, in which case it won't. BTW, we can use a more modern type notation, e.g. `list[
                dict[str, Any] | list[dict[str, Any]]]`.

MarleneKress79789: huh, i have never actually seen the later case happen i think. i will look into it

tkilias: The reason why you never encountered it is the following line

https://github.com/exasol/transformers-extension/blob/8466f7ffef22b8e93e8b8f6a6e4177808a35689a/exasol_transformers_extension/udfs/models/token_classification_udf.py#L69

ahsimb: Then, maybe changing the type hint is all we need to do.

MarleneKress79789: i updated the type hints, but also added an extra check cause i am not sure which format empty results would have. this way they are always dealt with

exasol_transformers_extension/udfs/models/token_classification_udf.py
```diff
@@ -159,7 +158,7 @@ def create_dataframes_from_predictions(
                                    f"Prediction results contain columns: {result_df_column_names}") from e
             else:
                 # if the result for an input is empty, just append an empty result df, and the input will be dropped during concatenation
-                result_df = pd.DataFrame(result)
+                result_df = pd.DataFrame({})
```
tkilias, line 161: the dataframe still needs the correct columns

MarleneKress79789: if the result is empty it does not have them either way. these results are later dropped before anything with the columns is done. this is basically just a placeholder to make sure we drop the right input together with the empty output.

tkilias (approved):

PR #301, opened 2025-03-06
Changed upload_model to have pyton api

### LINKED ISSUE #149: Create an API for uploading a model to BucketFS equivalent to the CLI upload_model.
At the moment the way to invoke this functionality programmatically is calling three functions like this :
```
# Create bucketfs location
bucketfs_location = create_bucketfs_location(bfs_service, host_name, bfs_port, bfs_use_https, bfs_user, bfs_password,
    bfs_bucket, bfs_dir)
# Upload the downloaded model files into bucketfs
upload_path = get_model_path(bfs_model_dir, model_name)
upload_model_files_to_bucketfs(model_dir, upload_path, bucketfs_location)
```
It would be good to combine this into a single call.

MarleneKress79789:
### All Submissions:

fixes #149

### DIFF
--- a/exasol_transformers_extension/upload_model.py
+++ b/exasol_transformers_extension/upload_model.py
@@ -1,8 +1,10 @@
 from __future__ import annotations
 from pathlib import Path
+from typing import Optional
 
 import click
 import transformers
+from exasol.bucketfs._path import PathLike
 
 from exasol.python_extension_common.cli.std_options import (
     StdTags, select_std_options, make_option_secret)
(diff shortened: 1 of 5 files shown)


exasol_transformers_extension/upload_model.py
```diff
@@ -32,21 +34,37 @@ def upload_model(**kwargs) -> None:
...
+    # create bucketfs location
+    bucketfs_location = create_bucketfs_location(**kwargs)
+
+    upload_model_to_bfs_location(kwargs[MODEL_NAME_ARG], kwargs[TASK_TYPE_ARG],
+                                 Path(kwargs[SUBDIR_ARG]), bucketfs_location, kwargs[TOKEN_ARG])
+
+
+def upload_model_to_bfs_location(model_name: str, task_type: str, subdir: Path,
+                                 bucketfs_location: PathLike, huggingface_token: Optional[str] = None) -> None:
+    """
```
ahsimb, line 46: Can you return the model file path rather than print it? The print is not necessarily what the caller wants to happen.

ahsimb (changes requested):

MarleneKress79789:
> It will be good to have a test for this function.

added unit tests.
integration test i would say [this](https://github.com/exasol/transformers-extension/blob/main/test/integration_tests/with_db/test_upload_model.py) is sufficient, it calls this function indirectly

exasol_transformers_extension/upload_model.py
MarleneKress79789: changed

test/unit/utils/test_upload_model_python_api.py
```diff
@@ -0,0 +1,27 @@
...
+from exasol_transformers_extension.upload_model import upload_model_to_bfs_location
+from utils.parameters import model_params
+
+
+def test_model_upload_python_api(tmpdir_factory):
+    #get specs for a valid huggingface model
+    model_specs = model_params.token_model_specs
+    sub_dir = Path("subdir")
+
+    mock_bucketfs_location = tmpdir_factory.mktemp("test_upload_python_api")
```
ahsimb, line 12: The name is not ideal. "mock" is something that pretends to be something else. When we patch a function we create a mock that pretends to be this function. In this case `mock_bucketfs_location` is a real directory, albeit a short-lived one.

ahsimb (approved):

PR #314, opened 2025-07-30
Bug/#285 fixed integration test asserts

MarleneKress79789:
### All Submissions:

### DIFF
--- a/doc/changes/unreleased.md
+++ b/doc/changes/unreleased.md
@@ -20,7 +20,7 @@ T.B.D
 
 ### Bugs
 
- n/a
+ - #285: Fixed logic in model output quality checks of udf integration test
 
 ### Dependency Updates
 
--- a/noxfile.py
+++ b/noxfile.py
@@ -74,7 +74,8 @@ def onprem_integration_tests(session):
         "-s",
         "--backend=onprem",
         "--itde-db-version=external",
-        "test/integration_tests/with_db",
+        #"test/integration_tests/with_db",
+        "test/integration_tests/with_db/udfs/test_sequence_classification_single_text_script.py"
     )
 
 
--- a/test/integration_tests/with_db/udfs/test_filling_mask_script.py
+++ b/test/integration_tests/with_db/udfs/test_filling_mask_script.py
@@ -45,15 +45,14 @@ def test_filling_mask_script(
     n_cols_result = len(input_data[0]) + (added_columns - removed_columns)
     assert len(result) == n_rows_result and len(result[0]) == n_cols_result
 
-    # lenient test for quality of results, will be replaced by deterministic test later
     results = [result[i][5] for i in range(len(result))]
 
-    # lenient test for quality of results.
+    # Lenient test for quality of results.
     # We do this by seeing if the result contains one of our predefined "acceptable_results".
-    # this check is only here to assure us the models output is not totally of kilter
+    # This check is only here to assure us the models output is not totally of kilter
     # (and crucially does not get worse with our changes over time),
     # and therefore we can assume model loading and execution is working correctly.
-    # a plan to make this check deterministic in the future exists.
+    # We to make this check deterministic in the future.
     acceptable_results = ["love", "miss", "want", "need"]
     number_accepted_results = 0
 
--- a/test/integration_tests/with_db/udfs/test_filling_mask_script.py
+++ b/test/integration_tests/with_db/udfs/test_filling_mask_script.py
@@ -47,6 +47,13 @@ def test_filling_mask_script(
 
     # lenient test for quality of results, will be replaced by deterministic test later
     results = [result[i][5] for i in range(len(result))]
+
+    # lenient test for quality of results.
+    # We do this by seeing if the result contains one of our predefined "acceptable_results".
+    # this check is only here to assure us the models output is not totally of kilter
+    # (and crucially does not get worse with our changes over time),
+    # and therefore we can assume model loading and execution is working correctly.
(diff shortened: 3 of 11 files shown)


test/integration_tests/utils/model_output_quality_checkers.py
```diff
@@ -0,0 +1,104 @@
...
+    label_index: int = 5,
+):
+    """
+    Lenient test for quality of results.
+    We do this by seeing if the result is one of our predefined "acceptable_results".
+    This check is only here to assure us the models output is not totally of kilter
+    (and crucially does not get worse with our changes over time),
+    and therefore we can assume model loading and execution is working correctly.
+    We plan to make this check deterministic in the future.
+    """
```
ahsimb, line 67: Same concerns as for the previous function.

test/integration_tests/utils/model_output_quality_checkers.py
```diff
@@ -0,0 +1,104 @@
...
+    (and crucially does not get worse with our changes over time),
+    and therefore we can assume model loading and execution is working correctly.
+    We to make this check deterministic in the future.
+    """
+    results = [
+        [result[i][label_index], result[i][label_index + 1]] for i in range(len(result))
+    ]
+    number_accepted_results = 0
+
+    for i in range(len(results)):
```
ahsimb, line 99: Please use Python :). E.g.
`number_accepted_results = sum(1 for res in results if res in acceptable_result_sets)`

test/integration_tests/utils/model_output_result_number_checker.py
```diff
@@ -0,0 +1,28 @@
+def assert_correct_number_of_results(
+    added_columns: int,
+    removed_columns: int,
+    input_data_row: tuple,
+    result: list,
+    n_rows_result: int,
+):
+    n_cols_result = len(input_data_row) + (added_columns - removed_columns)
+    assert len(result) == n_rows_result and len(result[0]) == n_cols_result, (
```
ahsimb, line 9: I would make a separate statement for checking the number of columns, like
`num_columns_ok = all(len(row) == n_cols_result for row in result)`
Then
`assert num_columns_ok and len(result) == n_rows_result`
Or, my personal preference is separate asserts - easier to see where it breaks and you can make the error messages more specific.

test/integration_tests/utils/model_output_quality_checkers.py
MarleneKress79789: this seems like a difference in personal preference to me

test/integration_tests/utils/model_output_result_number_checker.py
MarleneKress79789: torsten often prefers one assert statement. i am fine with both versions, i dont think it makes that much of a difference here

test/integration_tests/utils/model_output_result_number_checker.py
```diff
@@ -0,0 +1,28 @@
...
+
+def assert_correct_number_of_results_multiple_results_per_input(
+    added_columns: int,
+    removed_columns: int,
+    input_data_row: tuple,
+    result: list,
+    n_rows: int,
+):
+    n_cols_result = len(input_data_row) + (added_columns - removed_columns)
+    assert len(result) >= n_rows and len(result[0]) == n_cols_result, (
```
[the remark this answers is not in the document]
MarleneKress79789: i can if it is important to you, however i doubt it would make much of a difference in both lines of code and readability

test/integration_tests/utils/model_output_quality_checkers.py
```diff
@@ -0,0 +1,104 @@
...
+    --------------------------------------------------------------
+    low confidence        | bad result        |  acceptable
+    (result_score < 0.2)  |                   |
+
+    We only sum up acceptable results below, because we already know we
+    have the correct number of results from the other checks.
+    """
+
+    def contains(string, list):
+        return any(map(lambda x: x in string, list))
```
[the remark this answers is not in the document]
MarleneKress79789: this is kind of intentional. not all of our udfs have deterministic results. this is just a sanity check to make sure the results are not completely of the rails. so we assume if our results contain certain words or phrases, they are good enough. this needs to allow for other parts of the result string to be different. i tried to design the inputs and expected outputs for the tests in a way where we are able to define a "good enough" in a way that satisfies our needs.
as stated in the comments, there are plans under consideration which would make the outputs deterministic. but i have no control over when they will be executed.
anyway, even if this is not satisfactory to you, i don't think it is part of this ticket.

test/integration_tests/utils/model_output_quality_checkers.py
```diff
@@ -0,0 +1,104 @@
...
+):
+    """
+    Lenient test for quality of results.
+    We do this by seeing if the result is one of our predefined "acceptable_results".
+    This check is only here to assure us the models output is not totally of kilter
+    (and crucially does not get worse with our changes over time),
+    and therefore we can assume model loading and execution is working correctly.
+    We to make this check deterministic in the future.
+    """
+    results = [
```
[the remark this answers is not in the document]
ahsimb: Seriously?

test/integration_tests/utils/model_output_quality_checkers.py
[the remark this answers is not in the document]
MarleneKress79789: regardless, this is just code that was moved and is not part of this ticket. feel free to talk to pm or torsten if you find it important that we fix this, and if they agree we can do so in another ticket.

PR #368, opened 2026-02-13
 improve max new token in unit tests

### LINKED ISSUE #348: improve unit tests involving max_lenght/max_new_tokens
the unit tests currently do not make much sense, it would be better to use a function which utilizes max_lenght and input-text as side-effect for the pipeline-mock, instead of the current list which is used. this would allow us to actually check if the udf propagates the max_lenght param correctly.

MarleneKress79789:
### All Submissions:

* [ ] Have you updated the User/Developer Guide?
fixes #348


doc/changes/unreleased.md
```diff
@@ -39,11 +42,8 @@ T.B.D
 
 ## Refactorings
 
-* Updated urllib3 (2.5.0 -> 2.6.2)
```
kratz00: Not sure if relevant:

You've removed the entry mentioning the update to version 2.6.2, keept the 2.6.1 one and the lock file says 2.6.3 (see https://github.com/exasol/transformers-extension/blob/main/poetry.lock#L5090)

test/unit/udf_wrapper_params/ai_translate_extended/make_data_row_functions.py
```diff
@@ -78,15 +78,25 @@ def make_udf_output_for_one_input_row(
...
+            return Exception(
+                "Traceback mock_pipeline is throwing an error intentionally"
+            )
+        # remove the input prefix from the input
+        input_prefix, text = input_text.split(": ")
+        lang_marker = " übersetzt, "
+        if "French" in input_prefix:
+            lang_marker = " traduit, "
+        else:
+            lang_marker = " übersetzt, "
```
kratz00:
```suggestion

```

kratz00 (approved):

PR #382, opened 2026-04-16
add default param  transformations

### LINKED ISSUE #381: add remove-columns transformation
for udfs with default values we need to b able to remove columns after the prediction is done.

### LINKED ISSUE #383: add add_default_columns transformation

MarleneKress79789:
### All Submissions:

* [x] Have you updated the User/Developer Guide?

Fixes #383, Fixes #381


exasol_transformers_extension/udfs/models/transformation/add_default_columns.py
```diff
@@ -0,0 +1,76 @@
...
+)
+from exasol_transformers_extension.utils.load_local_model import LoadLocalModel
+
+
+class AddDefaultColumnsTransformation(Transformation):
+    """
+    Transformation adds default columns filled with default values to input dataframe.
+    Will overwrite existing columns with same name with default values.
+
+    If a column should be added which we do not have default values for, this column
```
tkilias, line 24: I think, this transformation should get the default values as input into the constructor. This way different UDFs can have different defaults.

exasol_transformers_extension/udfs/models/transformation/add_default_columns.py
```diff
@@ -0,0 +1,76 @@
...
+    and all columns after this column will be added, but will be filled with "None",
+    and an error will be added to "error_message"
+    """
+
+    def __init__(
+        self,
+        expected_input_columns: list[str],
+        new_columns: list[str],
+        removed_columns: list[str],
```
tkilias, line 33: Why make this a parameter, if it should be anyway empty

exasol_transformers_extension/udfs/models/transformation/remove_columns.py
```diff
@@ -0,0 +1,57 @@
...
+
+class RemoveColumnsTransformation(Transformation):
+    """
+    Transformation removes given columns from a dataframe
+    """
+
+    def __init__(
+        self,
+        expected_input_columns: list[str],
+        new_columns: list[str],
```
tkilias, line 24: Why do we need this if it should be empty

exasol_transformers_extension/udfs/models/transformation/add_default_columns.py
MarleneKress79789: will change

exasol_transformers_extension/udfs/models/transformation/add_default_columns.py
MarleneKress79789: i have them for consistency, because all transformations have them. but i can also remove unnecessary ones for all transformations

MarleneKress79789: i was wrong, will remove the unnecessary inputs

exasol_transformers_extension/udfs/models/transformation/add_default_columns.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def transform(
+        self, batch_df: DataFrame, model_loader: LoadLocalModel
+    ) -> Iterator[DataFrame]:
+        """
+        add default columns filled with default values to input dataframe.
+        """
+        for default_column in self.new_columns:
+
+            if default_column == "model_name":
```
tkilias: I wouldn't handle the model differently and actually also inject that via the default_values dict

MarleneKress79789: that would make it harder to add the "check if all default models are available on huggingface" test you wanted, if  we split up the model definitions accross udfs. as it is now the model is inferred from the udf-name, which still means every default udf can have a different model, but we have a central and single point of truth.

tkilias: The UDF definition would still fetch the model from the central place, but I don't want to hard code this place here.

MarleneKress79789: ok, will change

tkilias (approved):

